PR #1, opened 2018-10-10
fix cluster datasource(s) if no http/smtp credentials are configured

simonfuhrer:


JonKohler:
Ahha! Ok, so this is the case where you have HTTP Proxy and/or SMTP server configured, but they are unauthenticated, correct? Just want to make sure I'm understanding the intention. That should make sense (as most people do it that way), and if so, good catch!

simonfuhrer:
Yes correct Jon

PR #7, opened 2018-11-01
correcting example

olljanat:
Closes https://github.com/nutanix/terraform-provider-nutanix/issues/216

### DIFF
--- a/examples/main.tf
+++ b/examples/main.tf
@@ -184,7 +184,6 @@ resource "nutanix_virtual_machine" "demo-01-web" {
   num_vcpus_per_socket = 2
   num_sockets          = 1
   memory_size_mib      = 4096
-  power_state          = "ON"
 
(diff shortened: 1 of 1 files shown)


marinsalinas (approved):
LGTM!

Also, the image_uri is getting some errors when you are testing on GCP, because it can't reach the iso file.

@JonKohler, Do we have a public iso image to add to the example?

olljanat:
Is http://download.cirros-cloud.net/0.3.4/cirros-0.3.4-x86_64-disk.img actually same than which you have on example now?

JonKohler:
Opened https://github.com/terraform-providers/terraform-provider-nutanix/issues/9 to cover this, this will address when done #9

olljanat:
@JonKohler / @marinsalinas what is the plan with this project now? Why not even this kind of minor things gets merged?

PR #39, opened 2019-03-21
Remove kind as a argument in cluster_reference and metadata.

### LINKED ISSUE #27: hard code (kind) to be appropriate kind for each resource
Let's do this so it doesn't *need* a metadata group in the .tf definition file. We're going to make this implicit in a future API update, so hardcode should be fine.

We should do this for anything where the kind is 1:1, like this:

  cluster_reference = {
    kind = "cluster"
    uuid = "${data.nutanix_clusters.clusters.entities.0.metadata.uuid}"
  }

marinsalinas:
Closes #27 

The new way to reference a cluster is:

``` hcl

resource "nutanix_virtual_machine" "vm1" {
  name = "test-virtual-machine"
  cluster_uuid = "${data.nutanix_clusters.clusters.entities.0.metadata.uuid}"
  num_vcpus_per_socket = 1
  num_sockets                  = 1
  memory_size_mib          = 186

  disk_list = [{
    data_source_reference = [{
      kind = "image"
      uuid = "${nutanix_image.cirros-034-disk.id}"
    }]
    disk_size_mib = 44
  }]
}
```


paultyng (approved):
This looks ok to me, its going to be a major version bump / breaking change though for users, but I'm guessing that is intended in this case?

nutanix/data_source_nutanix_subnet_test.go
```diff
@@ -108,19 +105,14 @@ output "cluster" {
...
-
+	cluster_uuid = "${data.nutanix_clusters.clusters.entities.0.metadata.uuid}"
 	vlan_id = %d
 	subnet_type = "VLAN"
 
 	prefix_length = 24
 	default_gateway_ip = "192.168.0.1"
 	subnet_ip = "192.168.0.0"
-	#ip_config_pool_list_ranges = ["192.168.0.5", "192.168.0.100"]
+	ip_config_pool_list_ranges = ["192.168.0.10 192.168.0.100"]
```
paultyng, line 115: not sure I fully understand why this is just a single string here? is it that each IP is part of the range?

marinsalinas: It is because this argument could accept different IP ranges, the API receive those ranges as an array of strings, so, to follow the same semantic we set this argument like this. 

You can set another ip range like this.
```hcl
ip_config_pool_list_ranges = ["192.168.0.10 192.168.0.100", "192.168.0.150 192.168.0.180"]
````

We also added a validation in the schema.
https://github.com/terraform-providers/terraform-provider-nutanix/blob/dd96968d4e0a60ef4d49569fa77bdf49a42989c3/nutanix/resource_nutanix_subnet.go#L219

PR #45, opened 2019-04-04
Add Serial Port support in VM resource and data source

### LINKED ISSUE #25: Ability to add a serial port to a VM
Moved from https://github.com/nutanix/terraform-provider-nutanix/issues/159

The situation as reported by leonteq: 

> when using Ubuntu 16.04 cloud image, like the one you’d deploy in aws/azure/etc:
> - The image is built with enabled output on serial port; if the VM has no serial port, there’s a very long pause before it times out and moves on
> - We fixed it by removing the serial port dependency in the image
> - The other workaround is to manually create serial port on the VM and restart

Answer from Jon 

> We’ve got the ability to do this in the code already in the v3 API that will ship with AOS 5.9, which is due out later this year. When that hits, serial_port_list will be the array you’d call configure the serial_port, just like you’d do to configure a NIC with nic_list or DISK with disk_list
>  
> The particular feature that was tied too is unlikely to be backported to 5.6/5.8 (which BTW 5.8 is coming out this week), so we’ll need to wait for that API to have terraform hook into it.

This is similar to the issue about resizing disks, the API support is now there in 5.9+, and 5.10 as such, so we'll need to validate and make sure its documented correctly here.

marinsalinas:
closes #25

### DIFF
--- a/nutanix/resource_nutanix_virtual_machine.go
+++ b/nutanix/resource_nutanix_virtual_machine.go
@@ -1704,8 +1704,9 @@ func preFillResUpdateRequest(res *v3.VMResources, response *v3.VMIntentResponse)
 	}
 	res.NicList = nold
 
-	spl := make([]*v3.VMSerialPort, len(response.Status.Resources.SerialPortList))
+	var spl []*v3.VMSerialPort
 	if len(response.Status.Resources.SerialPortList) > 0 {
+		spl = make([]*v3.VMSerialPort, len(response.Status.Resources.SerialPortList))
 		for k, v := range response.Status.Resources.SerialPortList {
 			spl[k] = &v3.VMSerialPort{
 				Index:       v.Index,
@@ -1713,8 +1714,6 @@ func preFillResUpdateRequest(res *v3.VMResources, response *v3.VMIntentResponse)
 			}
 
 		}
-	} else {
-		spl = nil
 	}
 	res.SerialPortList = spl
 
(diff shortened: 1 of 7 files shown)


nutanix/resource_nutanix_virtual_machine.go
```diff
@@ -1652,6 +1704,20 @@ func preFillResUpdateRequest(res *v3.VMResources, response *v3.VMIntentResponse)
...
+			spl[k] = &v3.VMSerialPort{
+				Index:       v.Index,
+				IsConnected: v.IsConnected,
+			}
+
+		}
+	} else {
+		spl = nil
+	}
+	res.SerialPortList = spl
```
thetonymaster, line 1718: you can do this in one step

nutanix/resource_nutanix_virtual_machine.go
```diff
@@ -9,7 +9,7 @@ import (
 	"strings"
 	"time"
 
-	"github.com/terraform-providers/terraform-provider-nutanix/client/v3"
+	v3 "github.com/terraform-providers/terraform-provider-nutanix/client/v3"
```
thetonymaster, line 12: Isn't this import by default `v3` without the name? Why using a named import?

thetonymaster (changes requested):

nutanix/resource_nutanix_virtual_machine.go
marinsalinas: Is the way that go 1.12 version do the goimports

nutanix/resource_nutanix_virtual_machine.go
marinsalinas: Done @thetonymaster

thetonymaster (approved):

PR #51, opened 2019-04-18
Data Source for categories

### LINKED ISSUE #8: Datasources for categories
From @ranman166 on https://github.com/nutanix/terraform-provider-nutanix/issues/217

> We are looking to be able to reference categories without having them managed as resources.
> 
> For instance, we want to be able to create a bunch of category keys and later have users add values if needed and reference them within their vm creation without having users create or destroy the keys.
> 
> A win would be referencing them without having to deal with the "id" within the vm resource. We could do this through local variables in a module, but managing their "values" will get out of hand.

marinsalinas:
closes #8 

We added category_key data source where you can find the values of the category if it has al least one.

### DIFF
--- a/nutanix/data_source_nutanix_category_key.go
+++ b/nutanix/data_source_nutanix_category_key.go
@@ -1,0 +1,81 @@
+package nutanix
+
+import (
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	v3 "github.com/terraform-providers/terraform-provider-nutanix/client/v3"
+	"github.com/terraform-providers/terraform-provider-nutanix/utils"
+)
+
+func dataSourceNutanixCategoryKey() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceNutanixCategoryKeyRead,
+
+		Schema: map[string]*schema.Schema{
+			"system_defined": {
+				Type:     schema.TypeBool,
+				Computed: true,
+			},
+			"description": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"api_version": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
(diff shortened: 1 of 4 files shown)


nutanix/data_source_nutanix_category_key.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	d.Set("api_version", utils.StringValue(resp.APIVersion))
+	d.Set("name", utils.StringValue(resp.Name))
+	d.Set("description", utils.StringValue(resp.Description))
+	d.Set("system_defined", utils.BoolValue(resp.SystemDefined))
+
+	d.SetId(utils.StringValue(resp.Name))
+
+	list, err := conn.V3.ListCategoryValues(d.Get("name").(string), &v3.CategoryListMetadata{
```
thetonymaster, line 66: My only concern here is pagination, not really an issue here, but I think there is a limit on how much categories can be created, isn't it?

nutanix/data_source_nutanix_category_key.go
```diff
@@ -0,0 +1,81 @@
...
+		Kind:   utils.StringPtr("category"),
+		Length: utils.Int64Ptr(500),
+	})
```
thetonymaster, line 69: Also, this is kinda hard to read, I know it seems easier to send the struct as an argument instead of creating a variable, but it gives unnecessary length to the call

nutanix/data_source_nutanix_category_key.go
```diff
@@ -0,0 +1,81 @@
...
+
+	if err != nil {
```
thetonymaster, line 71: related to the comments above, we can move this if to the call above, instead of leaving it here alone

thetonymaster (approved):

nutanix/data_source_nutanix_category_key.go
marinsalinas: Yes, I was thinking about it as well, I already have a workaround on this but I think I'll put it in a next enhancement phase.

PR #80, opened 2019-10-01
Add missing API fields for structs

JRemitz:
I've noticed after working with the Nutanix `v3` [API](https://www.nutanix.dev/reference/prism_central/v3/api/) that some fields were missing.

Added:
1. `NumThreads` in [**VMResources**](https://www.nutanix.dev/reference/prism_central/v3/api/vms/postvms#request-body)
1. `DataSourceReference` in [**ImageResources**](https://www.nutanix.dev/reference/prism_central/v3/api/images/postimages#request-body) - closes #52 
1. `ShouldForceTranslate` in [**Metadata** - used all over](https://www.nutanix.dev/reference/prism_central/v3/api/images/postimages#request-body)
1. Fixed typo for `omitempty`

PR #81, opened 2019-10-01
Add support for mapstructure decode tag for reusability

JRemitz:
I would like to leverage the client v3 API structs for Packer objects to the Nutanix API server. This would require adding support for [mapstructure](https://github.com/mitchellh/mapstructure) based on [Packer's documentation](https://www.packer.io/docs/extending/custom-builders.html#the-quot-prepare-quot-method) to simplify the process of parsing the Packer config.

I currently have [**Packer PR 8175**](https://github.com/hashicorp/packer/pull/8175) open to add support for Nutanix - any help/insight to align projects would be greatly appreciated!

This includes changes from #80 along with corresponding `mapstructure` tags.


marinsalinas:
@JRemitz  could you update with the latest code from master?

JRemitz:
I can and will take a look at this in a bit.

JRemitz:
@marinsalinas - I've updated however I see that the linter is failing on some of the long lines due to the extended tags. What's the best way to break those down?

marinsalinas:
@JRemitz, for now, you can add this into https://github.com/terraform-providers/terraform-provider-nutanix/blob/master/.golangci.yml configuration. See: https://github.com/golangci/golangci-lint/blob/a85e1e1faaef7f2f6a091e9b93df6be09539a009/.golangci.example.yml#L207

JRemitz:
@marinsalinas - apologies for the delay but finally got around to "fixing" this. Please take a look and let me know what you think. Thanks!

marinsalinas (changes requested):
Just resolve conflicts with go.mod

JRemitz:
Merged - how about now, @marinsalinas ?

marinsalinas (approved):

PR #88, opened 2019-11-29
implement session based authentification

maxaudron:
This PR implements session based authentication as an alternative to using basic auth for each request.

In case session_auth is set to true (default false) `NewClient()` makes a call to `/users/me` with basic auth and saves the returned cookies into the the `Client` object. 

The `NewRequest()` function uses these cookies if the `Client` has them set else uses basic auth.

Current limitation is that sessions expire after 15 minutes, a graceful retry mechanism should be added.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -87,7 +87,7 @@ func NewClient(credentials *Credentials) (*Client, error) {
 
 	c := &Client{credentials, httpClient, baseURL, userAgent, nil, nil}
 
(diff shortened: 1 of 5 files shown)


marinsalinas (approved):

marinsalinas:
@maxaudron, Could you please update the documentation, to add an example of how to set session auth with terraform?

marinsalinas (approved):

Issue #87, opened 2019-11-15
1122 API requests only to list VMs


mritzmann:
We manage several hundred servers with Terraform. A `terraform plan` lasts longer than an hour, which is not very convenient. That's why we took a look at Terraform's DEBUG Log. 

```shell
env TF_LOG=DEBUG TF_LOG_PATH="/tmp/terraform.log" terraform plan
```

First, some interesting statistics: Terraform executes 1404 API requests.

```shell
$ grep -i "/api/nutanix/" /tmp/terraform.log | wc -l
1404
```

Of these, 1122 are only for listing VMs.

```shell
$ grep -i "/api/nutanix/v3/vms/list" /tmp/terraform.log | wc -l
1122
```

I don't know golang enough, but I tried to find out where all the requests come from.

1. There is a function named [`resourceNutanixVirtualMachineExists`](https://github.com/terraform-providers/terraform-provider-nutanix/blob/34a9c14c88956bce549deb60599602e2a5b224d2/nutanix/resource_nutanix_virtual_machine.go#L1416-#L1431). I assume this function is called once for each individual VM, to check whether a VM exists.
2. Function `resourceNutanixVirtualMachineExists` calls a function named [`ListAllVM`](https://github.com/terraform-providers/terraform-provider-nutanix/blob/34a9c14c88956bce549deb60599602e2a5b224d2/client/v3/v3_service.go#L847-L883).
3. Function `ListAllVM` calls a function named [`ListVM`](https://github.com/terraform-providers/terraform-provider-nutanix/blob/34a9c14c88956bce549deb60599602e2a5b224d2/client/v3/v3_service.go#L132-L144).
4. Function `ListVM` makes multiple API requests to `/api/nutanix/v3/vms/list` (multiple requests because of pagination).

 Summary:

- For each VM, the entire list of all VMs is retrieved each time.
- Because of pagination this leads to several requests per server
- A request takes 5-15 seconds depending on the load.

A terraform state with ~500 servers thus leads to ~2500 requests

```
500 Servers * (1x5 pagination) = 2500 Requests
2500 Requests * ~ 7s = 17500s
```

Maybe there's a good reason why you're doing this... From my point of view it can never scale that way.

**Expected behavior**
The list with all servers should not be fetched every time anew. A single time per `terraform plan` is sufficient. In this example, only five requests would be necessary instead of 2500.

**Logs**
I'll be happy to deliver more if necessary.

**Versions**
 - OS: Debian 8.11
[...]

PR #92, opened 2020-01-07
Reduce the amount of API requests and improvements

maxaudron:
* add filter to ListAll*() function
    The filter allows us to, e.g. filter by name and dramaticly reduce the
    number of API requests as there is very little need to paginate anymore

* improve error messages
* remove unnecessary check for existence of resource  
    nutanix allows creating the same resource with the same name multiple
    times so we should allow this too.
    This also save us from doing some expensive API calls.

* get VM by ID for resourceNutanixVirtualMachineExists
    we have the ID available so it is not needed to loop through all the VMs

* remove ClusterIntentResource in favor of ClusterIntentResponse  
    ClusterIntentResource was only used in two places and is the exact same
    struct as ClusterIntentResponse.

### DIFF
--- a/nutanix/resource_nutanix_image.go
+++ b/nutanix/resource_nutanix_image.go
@@ -290,7 +290,7 @@ func resourceNutanixImageCreate(d *schema.ResourceData, meta interface{}) error 
 	// Make request to the API
 	resp, err := conn.V3.CreateImage(request)
 	if err != nil {
-		return err
+		return fmt.Errorf("error creating Nutanix Image %s: %+v", utils.StringValue(spec.Name), err)
 	}
 
 	UUID := *resp.Metadata.UUID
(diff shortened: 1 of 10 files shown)


marinsalinas:
Hello @maxaudron does this close #87 ?

nutanix/resource_nutanix_image.go
```diff
@@ -279,19 +279,10 @@ func resourceNutanixImageCreate(d *schema.ResourceData, meta interface{}) error
 	request.Metadata = metadata
 	request.Spec = spec
 
-	imageUUID, err := resourceNutanixImageExists(conn, n.(string))
```
marinsalinas, line 290: Why did you remove this validation?

coderGo93:
Hello, @maxaudron , there's a conflict, can you resolve it?

nutanix/resource_nutanix_image.go
maxaudron: The API does allow images with the same name so i don't see a reason why the provider should change that.

Should the API change to not allow this, the creation will fail anyways.

PacoDw:
hello, @maxaudron could you do rebase to master branch plz to update the Travis file?

marinsalinas (approved):

marinsalinas (approved):

PR #103, opened 2020-03-23
Added preCheck function to verify that env variables were set

PacoDw:
We can use this function in all the test cases to verify if the environment variables were correctly set.

### DIFF
--- a/nutanix/provider_test.go
+++ b/nutanix/provider_test.go
@@ -33,6 +33,13 @@ func TestProvider_impl(t *testing.T) {
 }
 
 func testAccPreCheck(t *testing.T) {
+	if os.Getenv("NUTANIX_USERNAME") == "" ||
+		os.Getenv("NUTANIX_PASSWORD") == "" ||
+		os.Getenv("NUTANIX_INSECURE") == "" ||
+		os.Getenv("NUTANIX_PORT") == "" ||
+		os.Getenv("NUTANIX_ENDPOINT") == "" {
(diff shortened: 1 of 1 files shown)


nutanix/provider_test.go
```diff
@@ -33,6 +33,13 @@ func TestProvider_impl(t *testing.T) {
 }
 
 func testAccPreCheck(t *testing.T) {
+	if os.Getenv("NUTANIX_USERNAME") == "" ||
+		os.Getenv("NUTANIX_PASSWORD") == "" ||
+		os.Getenv("NUTANIX_INSECURE") == "" ||
```
coderGo93, line 38: is this a required?, otherwise all seems good to me

coderGo93 (approved):

nutanix/provider_test.go
PacoDw: This is only the function, but we can add this function in each acceptance test case to ensure/required those credentials. This avoids error for missing credencials :)

coderGo93: I understand, thank you 👍

marinsalinas (changes requested):
Can you add it to all acceptance test?

PacoDw:
@marinsalinas sure :) do you want those changes in this same PR?

marinsalinas:
Yes, please

PacoDw:
@marinsalinas I checked all the test and all of them have the function

marinsalinas (approved):
Thanks @PacoDw , LGTM, :shipit:

PR #106, opened 2020-03-25
bugfix device_properties in a disk_list

yannickstruyf3:
Fixed bug that is triggered when only device_type is passed with device_properties in a disk_list.


marinsalinas (approved):
LGTM! Thanks

PR #108, opened 2020-04-03
added cloudinit fix

yannickstruyf3:


marinsalinas (changes requested):
Plz resolve the linter issue, otherwise we can merge it.

PR #124, opened 2020-05-13
Fix #96: During a resource adjustment the VM is not shut down using ACPI

### LINKED ISSUE #96: During a resource adjustment the VM is not shut down using ACPI
**Describe the bug**
During a resource adjustment, the VM is restarted by Terraform. I noticed that the VM does not shut down cleanly with an ACPI event. Instead, the VM is turned off hard.

At least that's how I interpret the code:

https://github.com/terraform-providers/terraform-provider-nutanix/blob/34a9c14c88956bce549deb60599602e2a5b224d2/nutanix/resource_nutanix_virtual_machine.go#L1234-L1238

**Expected behavior**
* Terraform should send an ACPI shutdown and wait at least five minutes.
* The time to wait should be configurable.

**Logs**
I'll be happy to deliver more if necessary.

**Versions (please complete the following information):**
 - OS that is executing Terraform: Debian 8.11
 - Terraform: v0.11.14
 - Nutanix Cluster (Prism Element / AOS): 5.10.9.1 LTS
 - Nutanix Prism Central: 5.11.2.1
 - Terraform provider version: v1.0.2

**Additional context**

PacoDw:
Added:
- The new `power_state_mechanism_config` attribute was added, and we can use it as a following:

```hcl
resource "nutanix_virtual_machine" "vm1" {
  name         = "my-vm"
  cluster_uuid = local.cluster1

  num_vcpus_per_socket = 1
  num_sockets          = 1
  memory_size_mib      = 186

  power_state_mechanism_config {
    guest_transition_config {
      should_fail_on_script_failure = true
      enable_script_exec            = true
    }
    mechanism = "ACPI"
  }

}
```
Also, the following Terraform configuration can work with the old attributes that represented the above new configuration:
```hcl
resource "nutanix_virtual_machine" "vm1" {
  name         = "my-vm"
  cluster_uuid = local.cluster1

  num_vcpus_per_socket = 1
  num_sockets          = 1
  memory_size_mib      = 186

  should_fail_on_script_failure = true
  enable_script_exec            = true
  power_state_mechanism         = "ACPI"
}
```
But now they are deprecated, and we don't recomend using them due to inconsistency and bugs that were fixed with this new change.

Let me know if you have another comment or concern, thank you so much

closes #96


marinsalinas:
@PacoDw  could you please fix the conflicts?

nutanix/resource_nutanix_virtual_machine_test.go
```diff
@@ -257,7 +257,7 @@ func TestAccNutanixVirtualMachine_PowerStateMechanism(t *testing.T) {
 				ResourceName:            resourceName,
 				ImportState:             true,
 				ImportStateVerify:       true,
-				ImportStateVerifyIgnore: []string{"disk_list"},
+				ImportStateVerifyIgnore: []string{"power_state_mechanism_config", "power_state_mechanism"},
```
marinsalinas, line 260: ?Can't we add those values in the read function?

PacoDw: yes let change and check them

PacoDw: they are in the read function but they depend on a config to be set and they have the same attribute but with different structure, the `power_state_mechanism` will be deprecated, so I put them in ImportStateVerifyIgnore array due to a state error

marinsalinas (approved):
Just fix the conflicts otherwise LGTM

marinsalinas:
@yannickstruyf3  can you confirm if this is still needed since #96 is fixed with the timeout?

siddharth-nutanix:
As the corresponding issue is closed, we are closing this PR

Issue #89, opened 2019-12-13
Update Documentation for guest_customization_sysprep Attribute


8c8jseKa:
Currently, the documentation (at the below link) only reflects the `PREPARED` option for the `install_type:` guest_customization_sysprep attribute. Please also reference that the `FRESH` option is available, as this can affect if an `unattend.xml` named-file is generated (`PREPARED`) or an `autounattend.xml` named-file is generated (`FRESH`).

[https://www.terraform.io/docs/providers/nutanix/r/virtual_machine.html](https://www.terraform.io/docs/providers/nutanix/r/virtual_machine.html)

coderGo93:
Hello @8c8jseKa , I updated the doc with `FRESH` option as a valid value for `install_type` in this [PR](https://github.com/terraform-providers/terraform-provider-nutanix/pull/133)

PR #133, opened 2020-05-28
docs: added more info about Sysprep for install type

coderGo93:
Fix to #89 
Added more info for Sysprep install type with FRESH data in doc


marinsalinas (approved):
LGTM!

Issue #83, opened 2019-10-23
Failed to perform NGT ISO mount operation with error code: kNoFreeCdRomSlot


mlrtime:
Describe the bug
Terraform error with nutanix plugin

Expected behavior
plan should execute normally

Logs
data.nutanix_image.centos: Refreshing state...
data.nutanix_clusters.available: Refreshing state...
data.nutanix_subnet.primary_ny5: Refreshing state...
data.nutanix_cluster.cluster: Refreshing state...
nutanix_virtual_machine.exd360[0]: Refreshing state... [id=f6861630-6f0f-44fe-8792-679cf4c041fc]
nutanix_virtual_machine.exd360[1]: Refreshing state... [id=a795c8e8-d6f8-468f-8afa-64a2b41f5101]

Error: error reading Virtual Machine f6861630-6f0f-44fe-8792-679cf4c041fc: error: {
  "message_list": [
    {
      "message": "Failed to perform NGT ISO mount operation with error code: kNoFreeCdRomSlot",
      "reason": "INTERNAL_ERROR"
    }
  ],
  "state": "ERROR"
}

Versions (please complete the following information):

OS: RHEL 7
Terraform 0.12.9
Nutanix Terraform Provider v1.0.2

Additional context

We tried mounting the NGT through prism central, ever since then we are unable to plan a state with two vms.  We are getting this error.  I removed all cdrom devices through prism.

      "message": "Failed to perform NGT ISO mount operation with error code: kNoFreeCdRomSlot",
      "reason": "INTERNAL_ERROR"

PR #134, opened 2020-05-28
fix #112: Improve error handling on incorrect API calls

### LINKED ISSUE #112: Improve error handling on incorrect API calls
When performing 'incorrect' API calls via the terraform provider, the "status" field on the API is being set to "error". Trying to shrink disks will cause this state.

 This state makes it impossible to use Terraform afterwards. 

The only way of getting the VM out of error state is the following:
 - Retrieve the UUID of the virtual machine (should be in your statefile)
 - Perform a GET  request on https://<< Prism IP >>:9440/api/nutanix/v3/vms/<< VM UUID >>
 - Take the output, remove the “status” key from the JSON
 - Perform a PUT request using the retrieved payload on https://<< Prism IP >>:9440/api/nutanix/v3/vms/<< VM UUID >>

Improving error handling should resolve this issue.

PacoDw:
Validation for Delete Method used as a filter for all the server responses was removed, allowing us to pass all the responses with code 200 despite if they contain any internal error so it now allows us to remove resources in TF when they crash.

Let me know if you have a comment or concern, thanks

closes #112 #83

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -232,7 +232,7 @@ func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) error
 
 // CheckResponse checks errors if exist errors in request
 func CheckResponse(r *http.Response) error {
-	if c := r.StatusCode; c >= 200 && c <= 299 && r.Request.Method == http.MethodDelete {
+	if c := r.StatusCode; c >= 200 && c <= 299 {
 		return nil
 	}
 


marinsalinas:
LGTM just keep in mind the following:


> Since this change impacts all the requests, I’d recommend running all the testacc to see if there is no regressions, because I think the rest of the function is hardly invoked since it allows all 200 <= 299 responses, so I think the rest of the function needs to be refactored.

marinsalinas (approved):

Issue #86, opened 2019-11-06
CPU and RAM change does not restart VM automatically


mritzmann:
**Describe the bug**
My guest system does not support Hotplug out of the box, so after CPU and RAM adjustment a restart is required. Terraform automatically restarts the affected VM when the size is reduced, but not when the size is increased. This feature was added in PR #75.

**Expected behavior**
Terraform should restart the affected VM. Or alternatively provide an option to configure this behavior.

**Logs**
I'll be happy to deliver more if necessary.

**Versions (please complete the following information):**
 - OS: Debian 8.11
 - Terraform: v0.11.14
 - Nutanix Cluster (Prism Element / AOS): Version 5.10.6 LTS
 - Nutanix Prism Central: 5.10.6
 - Terraform provider version: v1.0.2, installed with `provider "nutanix"`

**Additional context**
I created this issue as a bug, because the behavior has changed, without the possibility to keep the old behavior. So I can't enlarge VMs via Terraform anymore. Or is there a configuration option for it?

mritzmann:
Build v1.0.2 without Hotplug:

```bash
docker run -it golang:1.12 /bin/bash

git checkout -b patch-revert 50863de314994937b0cf822ee63bf11e9c888ca8
git revert 7123f520971114e70accb598347bbbde871f602b
git revert 1ba8002fdbdb6202000c66496adebde451dda5c2
git revert 120d472d33c998e68b8e723a287b1e908babb4ab
git revert 1cc9c6db1124c9519aa1d8c1dff0affcde4f971c
git revert f61aa7364b0c15aa069e82846e5565771141a1db

go get -u github.com/golang/dep/cmd/dep

env CGO_ENABLED=0 GOOS=linux GOARCH=amd64 make build
```

mritzmann:
From NTNX Case # 00669213:

> In that case the only workaround to Guest VMs that does not support hot add as of now would to use “power_state” attribute (https://www.terraform.io/docs/providers/nutanix/d/virtual_machine.html#power_state).
> 
> This involves a two step process though. First would be to update the Terraform file with increased resources to what you need and also add the "power_state" parameter with value "OFF" and apply the changes. This would update the resources and power off the VMs. 
> 
> The second step would be to update the same file or have a separate file with "power_state" set to "ON" which will power on the powered off VMs.
[...]

Issue #79, opened 2019-09-23
Hotplugging CPU and RAM


bt-josh:
**Describe the bug**
Hotplugging for CPU and RAM appears to have been enabled in #75 however VMs are still rebooting after increasing either `num_vcpus_per_socket` or `memory_size_mib`.

**Expected behavior**
Number of CPUs or Memory should increase without a system restart.

**Logs**
N/A

**Versions (please complete the following information):**
 - OS: RHEL 7
 - Terraform 0.12.8
 - Connecting to: Prism Element 5.9
 - Not connecting to Prism Central
 - Nutanix Terraform Provider v1.0.2

**Additional context**
We don't currently use the Prism Central for our Terraform connections, we connect directly to the Prism Element instead.

Jorge-Holgado:
Good morning,
I didn't patch tf 0.12. Don't know if @marinsalinas  has ported the change to 0.12 branch.
Maybe he can solve our doubts.

bt-josh:
The TF 0.12 code was merged into master as part of the same provider release (v1.0.2) as the hotplug code, so I don't think there's a discrete 0.12 branch anymore.

mritzmann:
I found this commit from @Jorge-Holgado and @marinsalinas: https://github.com/terraform-providers/terraform-provider-nutanix/commit/f61aa7364b0c15aa069e82846e5565771141a1db

```
Patch to allow some changes to be hotplug

Options that will be hot-plug (always):
  * name
  * description

Options hot-plug just when increasing it:
  * num_vcpus_per_socket
  * num_sockets
  * memory_size_mib
```

In my case, however, I explicitly want a reboot, since my guest system does not support hotplug. Is there a way to force a restart by terraform? Is there a way to disable hot-plug?

coderGo93:
Hello @bt-josh  , I made [PR](https://github.com/terraform-providers/terraform-provider-nutanix/pull/136) with branch `fix-86/79` , a new parameter to force reboot for Cpus/memory changes so in your case it shouldn't reboot unless asked, so if it's possible you can try it and see if it's ok, please?

coderGo93:
Hello, I'm going to close because there are not responses from the issuer and it has been added a parameter for this case and merged to master and applies for #86 too.

PR #136, opened 2020-05-29
Added new parameter for VM use_hot_add

coderGo93:
Fix/enhancement for #86 and #79 

Added a new parameter called `use_hot_Add` in VM, for now it only applies when there a change for `num_vcpus_per_socket` or `memory_size_mib`, if it's specified as false, is going to force a reboot where there's change for `num_vcpus_per_socket` or `memory_size_mib` otherwise not reboot.

### DIFF
--- a/nutanix/resource_nutanix_virtual_machine.go
+++ b/nutanix/resource_nutanix_virtual_machine.go
@@ -279,6 +279,11 @@ func resourceNutanixVirtualMachine() *schema.Resource {
 
 			// RESOURCES ARGUMENTS
 
+			"use_hot_add": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
 			"num_vnuma_nodes": {
 				Type:     schema.TypeInt,
 				Optional: true,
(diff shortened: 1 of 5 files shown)


nutanix/resource_nutanix_virtual_machine.go
```diff
@@ -917,6 +922,9 @@ func resourceNutanixVirtualMachineRead(d *schema.ResourceData, meta interface{})
 	if err := d.Set("parent_reference", flattenReferenceValues(resp.Status.Resources.ParentReference)); err != nil {
 		return fmt.Errorf("error setting parent_reference for Virtual Machine %s: %s", d.Id(), err)
 	}
+	if err := d.Set("use_hot_add", d.Get("use_hot_add")); err != nil {
+		return fmt.Errorf("error setting use_hot_add for Virtual Machine %s: %s", d.Id(), err)
```
PacoDw, line 926: Why did you use d.Get() func to retrieve the use_hot_add value? Does the server not return that value?

PacoDw (changes requested):

nutanix/resource_nutanix_virtual_machine.go
coderGo93: That's correct, this parameter is made to force reboot using logic, there are not api for that case.

PacoDw: okay make sense

PacoDw (approved):
LGMT

Issue #44, opened 2019-04-02
cannot unmarshal string into GO struct field MessageResource.details of the type map [string]interface{}


ghost:
Getting the following error when setting up a vm on nutanix:
cannot unmarshal string into GO struct field MessageResource.details of the type map [string]interface{}

Please let me know if anyone has seen this error and what may be the cause

marinsalinas:
@mkapooretg Can you share your configuration file and Prism Central version as well?

titilambert:
On my side: 
```
Nutanix: Version 5.10.3
NCC Version: ncc-3.7.0.2
```

I just have


```
provider "nutanix" {
  username = "USERNAME"
  password = "PASSWORD"
  endpoint = "HOSTADDRESS"
  insecure = true
  port     = 9440
}

data "nutanix_clusters" "clusters" {
}
```

I tried:
* terraform v0.11.13 + terraform-nutanix 1.0
* terraform v0.11.11 + terraform-nutanix 1.0
* terraform v0.11.8 + terraform-nutanix 1.0
* terraform v0.11.13 + terraform-nutanix HEAD
* terraform v0.11.11 + terraform-nutanix HEAD
* terraform v0.11.8 + terraform-nutanix HEAD

Maybe it's because I don't have access to clusters ressources ?

marinsalinas:
@mkapooretg : can I have the crash.log?

rdobrynski:
I too seem to be experiencing this issue on terraform 0.11.14 and PC version 5.10.4
terraform plan works, but apply breaks.

ghost:
Check the login information. Make sure that is correct.


Thanks,
Maneesh Kapoor | Essextec
Manager Professional Services
P: 929-371-3648<tel:929-371-3648> | C: 929-341-0431<tel:929-341-0431>
mkapoor@essextec.com<mailto:eherbst@essextec.com>
Website<http://www.essextec.com/> | LinkedIn <https://www.linkedin.com/company/essextec> | Twitter<https://twitter.com/essextec>
 [id:image001.png@01D50CDC.BCD63130]
Cloud | Cognitive | Cybersecurity

On Jun 5, 2019, at 6:09 PM, rdobrynski <notifications@github.com<mailto:notifications@github.com>> wrote:


I too seem to be experiencing this issue on terraform 0.11.14 and PC version 5.10.4
terraform plan works, but apply breaks.

rdobrynski:
[...]

PR #148, opened 2020-06-30
provide better error reporting when invalid nutanix credentials are used

scott-grimes:
Related: #44 

Nutanix returns a non-json response with code 401 when invalid credentials are provided. Instead of printing the rather obscure message 

`cannot unmarshal string into GO struct field MessageResource.details of the type map [string]interface{}`

we will now print 

`Invalid Nutanix Credentials`


marinsalinas (approved):
LGTM

PR #156, opened 2020-07-06
Improvement to create images using 'data_source_reference'

saratkumar-yb:
* Improvement to create images using 'data_source_reference'


marinsalinas:
@kumarsarath588 can you add a test case and website documentation for this feature?

siddharth-nutanix:
@kumarsarath588 Can you rebase your PR with master and change the type from map to list with max items as 1 as Maps doesn't support inner nested schema ? 

Another option could be to use just TypeMap with 
`&schema.Schema {
Type: schema.TypeString,
}`
Check https://www.terraform.io/plugin/sdkv2/schemas/schema-types#typemap for more details.

nutanix/resource_nutanix_image.go
```diff
@@ -175,10 +195,11 @@ func resourceNutanixImageCreate(ctx context.Context, d *schema.ResourceData, met
 
 	_, iok := d.GetOk("source_uri")
 	_, pok := d.GetOk("source_path")
+	_, dsr := d.GetOk("data_source_reference")
 
-	// if both path and uri are provided, return an error
-	if iok && pok {
-		return diag.Errorf("both source_uri and source_path provided")
+	// if three path, uri, dsr are provided, return an error
+	if iok && pok && dsr {
```
siddharth-nutanix, line 201: What if only two values are provided ?
Then also we should raise error.

nutanix/resource_nutanix_image.go
```diff
@@ -139,6 +139,26 @@ func resourceNutanixImage() *schema.Resource {
 				Optional: true,
 				Computed: true,
 			},
+			"data_source_reference": {
+				Type:     schema.TypeList,
+				MaxItems: 1,
+				Optional: true,
+				Computed: true,
```
siddharth-nutanix, line 146: If the field is marked as "Computed" then we should ideally read its value in resourceNutanixImageRead().
It calls out to https://www.nutanix.dev/api_references/prism-central-v3/#/b3A6MjU1ODc1MTE-get-a-existing-image API. 
We can set this value in read function as well.

siddharth-nutanix (changes requested):

GullapalliAkhil:
This enhancement has been covered as part of this PR https://github.com/nutanix/terraform-provider-nutanix/pull/837, closing this.

PR #160, opened 2020-08-05
Removed volume_group documentations

PacoDw:
- Since the volume groups resource is not part of the Nutanix provider, it was removed

### DIFF
--- a/website/nutanix.erb
+++ b/website/nutanix.erb
@@ -1,71 +1,75 @@
 <% wrap_layout :inner do %>
-    <% content_for :sidebar do %>
-        <div class="docs-sidebar hidden-print affix-top" role="complementary">
-            <ul class="nav docs-sidenav">
-                <li<%= sidebar_current("docs-home") %>>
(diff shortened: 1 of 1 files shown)


coderGo93 (approved):
LGTM

marinsalinas (changes requested):
I think we can keep those files and just remove the link over this: https://github.com/terraform-providers/terraform-provider-nutanix/blob/master/website/nutanix.erb#L19

PacoDw:
Thanks, @marinsalinas it's better 👍

website/nutanix.erb
```diff
@@ -1,71 +1,75 @@
...
-        </div>
-    <% end %>
-
-    <%= yield %>
+    </ul>
+</div>
 <% end %>
+
+<%= yield %>
+<% end %>
```
marinsalinas, line 75: just add trailing line please :)

marinsalinas (approved):

PR #175, opened 2020-10-28
Access control policies

coderGo93:
Added Resource, datasource(s), tests, docs of Access Control Policies.

The acceptance tests has been passed.

Source: https://www.nutanix.dev/reference/prism_central/v3/api/access-control-policies/


nutanix/resource_access_control_policy.go
```diff
@@ -0,0 +1,755 @@
...
+			},
+			"description": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+			"metadata": {
+				Type:     schema.TypeMap,
+				Computed: true,
+				Elem: &schema.Resource{
```
marinsalinas: We can use it as TypeMap since it is only computed, but having Elem on TypeMap will raise an error, there is a two options:
- Leave is a Map and removes Elem.
- if we need to have all the attributes explicit we have to change to TypeList. with MaxValue=1,


 any thoughts? @yannickstruyf3

nutanix/resource_access_control_policy.go
```diff
@@ -0,0 +1,755 @@
...
+							Type:     schema.TypeString,
+							Optional: true,
+							Computed: true,
+						},
+					},
+				},
+			},
+			"categories": categoriesSchema(),
+			"owner_reference": {
+				Type:     schema.TypeMap,
```
marinsalinas: Change to TypeList with MaxValue=1 since this value is optional and computed.

nutanix/resource_access_control_policy.go
```diff
@@ -0,0 +1,755 @@
...
+						},
+						"name": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
+					},
+				},
+			},
+			"project_reference": {
+				Type:     schema.TypeMap,
```
marinsalinas: Same, TypeList with MaxValue=1

nutanix/resource_access_control_policy.go
```diff
@@ -0,0 +1,755 @@
...
+			"name": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+			"state": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"user_reference_list": {
```
marinsalinas: Just make sure if the subsequent apply doesn't show plan diff.

nutanix/resource_access_control_policy.go
```diff
@@ -0,0 +1,755 @@
...
+				},
+			},
+			"filter_list": {
+				Type:     schema.TypeList,
+				MaxItems: 1,
+				Optional: true,
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"context_list": {
```
marinsalinas: Can you move to the root since `filter_list` just have this argument so that can avoid unnecessary nesting complexity in the terraform schema,   we can change `context_list ` to `context_filter_list`

marinsalinas (changes requested):
Taken a look into the Schema and I left some comments.

Also, can you address the linting error as well?

nutanix/resource_access_control_policy.go
coderGo93: can you explain more?, I haven't had issues with that parameter for now.

nutanix/resource_access_control_policy.go
coderGo93: done, thank you marin

nutanix/resource_access_control_policy.go
coderGo93: done, thank you marin

nutanix/resource_access_control_policy.go
coderGo93: done but I changed to `filter_context_list` instead of `context_filter_list` if that's ok for you, marin

nutanix/resource_access_control_policy.go
marinsalinas: yeah sometimes having mixed required/optional/computed inside TypSet without hash function results with unwanted state diff in the second plan, just I want to make just the second plan have diff when setting these values.

nutanix/resource_access_control_policy.go
marinsalinas: Screen Shot 2020-11-02 at 13 50 20

 as you can see the documentation describes this parameter as context_filters_list so I think in order to be coherent with the API naming, I'd say change it to `context_filter_list`

coderGo93: understood, thank you marin

yannickstruyf3:
@coderGo93 Can we rename the files so that they match the others:
resource_access_control_policy.go => resource_nutanix_access_control_policy.go
resource_access_control_policy_test.go => resource_nutanix_access_control_policy.go

nutanix/resource_nutanix_project.go
```diff
@@ -306,7 +306,7 @@ func resourceNutanixProject() *schema.Resource {
 			"api_version": {
```
marinsalinas, line 307: I think this value should be computed and pass it hardoced in the APi requests similar than other resources.

marinsalinas:
@yannickstruyf3, next time, can you create another branch with your changes and make a PR against this branch in order to avoid conflicts if @coderGo93  is working on the same branch as well? I think this can avoid conflicts or unwanted changes.

yannickstruyf3:
@coderGo93 Modified the code for `resource_domain` to support maxItems 1 and to make it optional (in case no quotas are required)  
#183

marinsalinas (approved):

marinsalinas (approved):

PR #183, opened 2020-11-17
made resource_domain optional -> quotas are also optional in projects

yannickstruyf3:
Resource domain in projects should be optional -> It is possible create a project without quotas


marinsalinas (approved):
LGTM!

PR #186, opened 2020-11-17
Bugfix/acp user group typeset

yannickstruyf3:
correct bug where user_reference_list and user_group_reference_list is cast as list of interfaces but it is a typeset


marinsalinas:
Can you add this change on the base branch as well (`access_policies`) to have it as well?

yannickstruyf3:
> Can you add this change on the base branch as well (`access_policies`) to have it as well?

@marinsalinas  #187

PR #222, opened 2020-12-15
Add testacc for karbon resources and data sources

marinsalinas:


yannickstruyf3:
@marinsalinas Can we check more attributes (name, deployment_type, kubeapi_server_ipv4_address,...) in the acc tests?

yannickstruyf3:
Can we also add a test for a different CNI? This be done by changing "flannel" to Calico

yannickstruyf3:
Scale-up/down should also be part of the test cases. 
Changing the local.amount_of_workers parameter to a higher value will trigger scale up. Setting a lower value will trigger scale down.

PR #227, opened 2021-01-07
Bugfix/remove ide3 dependency

yannickstruyf3:


nutanix/structure.go
```diff
@@ -103,7 +104,113 @@ func flattenNicList(nics []*v3.VMNic) []map[string]interface{} {
...
+	}
+	if disk.DeviceProperties.DiskAddress.DeviceIndex == nil {
+		return nil, fmt.Errorf("device index was nil for disk")
+	}
+	diskIndex := *disk.DeviceProperties.DiskAddress.DeviceIndex
+	return &diskIndex, nil
+}
+
+func flattenDiskListFilterCloudInit(d *schema.ResourceData, disks []*v3.VMDisk) ([]map[string]interface{}, error) {
+	//todo check if guestcust is passed -> if it is not passed, just continue without searching for cloud-init uuid
```
marinsalinas, line 137: can we do something like d.GetOK(guestCustom...)? to check if tha value is set?

Also can we parse the cloud init script to check if actually the cd room is creating by the scrip?

nutanix/structure.go
```diff
@@ -103,7 +103,98 @@ func flattenNicList(nics []*v3.VMNic) []map[string]interface{} {
...
+		userCdromDiskList := GetCdromDiskList(expandedOrgDiskList)
+		for _, eDisk := range disks {
+			//if existing disk is not CD-rom, append it to the list and continue
+			if !isCdromDisk(eDisk) {
+				filteredDiskList = append(filteredDiskList, eDisk)
+				continue
+			} else {
+				//Get existing CDrom device Index
+
+				eDiskIndexP, err := getDeviceIndexForDisk(eDisk) //*eDisk.DeviceProperties.DiskAddress.DeviceIndex
```
marinsalinas, line 158: can you remove this line if it is not needed?

marinsalinas (changes requested):

marinsalinas (approved):
LGTM!

PR #231, opened 2021-01-15
removed project_ref

yannickstruyf3:


marinsalinas (approved):
Is this not needed in the current version of prism as well? 

I want to make sure both versions will work.

PR #234, opened 2021-01-19
Task completion project

coderGo93:
- Added delete response in service for projects
- Added waiting for task for update and delete projects


marinsalinas (approved):
The code LGTM! @yannickstruyf3  can you test to check if the behavior is the expected?

yannickstruyf3:
> The code LGTM! @yannickstruyf3 can you test to check if the behavior is the expected?

Tested and verified

yannickstruyf3 (approved):

PR #252, opened 2021-02-06
fix: add hotPlug change validation on memory_size_mib and num_socket …

marinsalinas:
…due regression


marinsalinas:
@yannickstruyf3 and @gabybeitler  can we merge this?

Issue #250, opened 2021-02-04
Hardcoded timeout causes VM build to fail


gfenn-newbury:
### Nutanix Cluster Information
Please provide the version of:
 - Nutanix Cluster (Prism Element / AOS): 5.18.1.2
 - Nutanix Prism Central: 2020.9.0.1
 - Provider Version: 1.2.0

### Terraform Version
0.14.4
### Affected Resource(s)
* nutanix_virtual_machine

### Expected Behavior
VM built after running remote-execs, which can take over 1 minute
### Actual Behavior
```
Error: error waiting for vm (18dee129-8b0d-4292-8304-9bdb14ebfaa7) to create: timeout while waiting for state to become 'SUCCEEDED' (last state: 'RUNNING', timeout: 1m0s)
  on .terraform/modules/test_vm/nutanix.tf line 11, in resource "nutanix_virtual_machine" "vm":
  11: resource "nutanix_virtual_machine" "vm" {
```
### Steps to Reproduce
Build VM on 1.2.0 which takes longer than 1 minute to build

~~This bug is cause by the following hard-coded timeouts with the nutanix_virtual_machine resource:~~

~~https://github.com/nutanix/terraform-provider-nutanix/blob/master/nutanix/resource_nutanix_virtual_machine.go#L21-L27~~

~~https://github.com/nutanix/terraform-provider-nutanix/blob/master/nutanix/resource_nutanix_virtual_machine.go#L884-L895~~

**Edit 2020-02-05: Not caused by the above code**

This causes a VM to fail its build if it continues for longer than 1 minute. ~~**There is no way to override this timeout**~~ Timeout can be overridden using `wait_timeout` in the provider block. However, this sets the timeout globally for the provider on all resources and all actions, and doesn't use [Hashicorp's documented method allowing a timeout block for individual resources](https://www.terraform.io/docs/language/resources/syntax.html#operation-timeouts).

~~The pull request which caused this issue is: https://github.com/nutanix/terraform-provider-nutanix/pull/101~~

gabybeitler:
There is a wait_timeout parameter in the provider. Could you use that parameter in your TF code and see if that accomplished what you need?  An example is below.

provider "nutanix" {
  wait_timeout = 2
  insecure     = true
  port         = 9440
}

gfenn-newbury:
Thanks, I'll give it a try and report back. I assume `wait_timeout` is in minutes?

[...]

PR #253, opened 2021-02-06
Feature/update documentation

gfenn-newbury:
* Updated documentation index page to be clearer on the arguments and what they do
* Added arguments which had missing or partially missing documentation
* Made the layout more standardised in comparison to other Terraform providers
* Added note about the `wait_timeout` argument being global, and [Hashicorp's operations timeouts](https://www.terraform.io/docs/language/resources/syntax.html#operation-timeouts) block not being supported, as this was the cause of issue #250


marinsalinas (approved):
LGTM! Thanks for the contribution!

PR #257, opened 2021-02-11
chore: fix index documentation provider's description

marinsalinas:
Quick fix, just to make sure the provider only works with prism central.


gfenn-newbury:
Has something changed recently? We're successfully using the provider with Prism Elements, and have also used it with Prism Central.

PR #261, opened 2021-02-15
refactor: use ListAll function to get all the result from the specifi…

marinsalinas:
…c resource


gabybeitler (approved):

yannickstruyf3 (commented):
Can we also add tests cases for the other data sources? Similar to what has been done for subnets?

marinsalinas:
> Can we also add tests cases for the other data sources? Similar to what has been done for subnets?

Actually, I would like to remove that subnet data source test case since this will cause conflicts when running all the tests together due to the VLAN id.

I think we can figure out how to test this better, in the next phase, since we need to figure out how to avoid conflicts with creating more than 20 resources and also avoid name conflicts with other resources from other tests cases.

yannickstruyf3 (approved):

Issue #269, opened 2021-04-01
unmarshalling error response invalid character ‘O’ looking for beginning of value


hkusanic:
**Description:**
This looks like Nutanix Terraform provider issue.

I am using following setup:
```
terraform {
    required_version = ">= 0.14"
      
    required_providers {
        nutanix = {
        source  = "nutanix/nutanix"        
        version = "~> 1.2"
        }
    }

    backend "local" {
       
    }
}
```
Error is always in the following form, but happens on different places in the terraform code:
```"unmarshalling error response invalid character ‘O’ looking for beginning of value"```

This issue is occasionally breaking our release pipeline, but oddly enough if you re-run the pipeline it will start to work without any code change in the codebase or in the release pipeline.

Bellow you can find logs from the 3 consecutive runs from the Azure DevOps pipeline (run 1-3) with the Terraform apply task.
Underneath, you can find Terraform files where error was reported

Again, after first two runs that failed, third one worked just fine without any errors. 
Mind you, same code, same artifacts, no changes.

**What I've tried already:**
I used Terraform validate to fix all the warnings in regards to Terraform variable syntax.

I’ve also seen following thread in slack but suggested workarounds didn't work for me.
https://app.slack.com/client/T0252CLM8/search/search-eba41109-7265-4d3e-9f86-5d690868278a/thread/C8E8RABB4-1549301669.049000 [app.slack.com]

**Expected result:**
There are no errors

**Logs:**
[run1.txt](https://github.com/nutanix/terraform-provider-nutanix/files/6241850/run1.txt)
[run2.txt](https://github.com/nutanix/terraform-provider-nutanix/files/6241851/run2.txt)
[run3.txt](https://github.com/nutanix/terraform-provider-nutanix/files/6241852/run3.txt)

**Terraform files:**
[centos001_main.tf.txt](https://github.com/nutanix/terraform-provider-nutanix/files/6241853/centos001_main.tf.txt)
[centos001_win2019_main.tf.txt](https://github.com/nutanix/terraform-provider-nutanix/files/6241854/centos001_win2019_main.tf.txt)
[ubuntu001_main.tf.txt](https://github.com/nutanix/terraform-provider-nutanix/files/6241855/ubuntu001_main.tf.txt)

yannickstruyf3:
Can you run your code in TRACE mode (https://www.terraform.io/docs/internals/debugging.html) and provide these logs? 
[...]

PR #270, opened 2021-04-01
modified unmarshal error to also return response body

yannickstruyf3:
Provides more verbose error message for issue #269


yannickstruyf3:
@gabybeitler @marinsalinas Still to be tested

marinsalinas (approved):
LGTM! Nice catch!

PR #280, opened 2021-05-25
Update karbon documentation for active_passive_config

pmiles:
Update Karbon documentation for active_passive_config setting


gabybeitler (changes requested):
Please remove "(which you should for production)"

PR #309, opened 2021-11-25
Skipping state refresh for ngt credentials while reading nutanix_vm.

siddharth-nutanix:
Behaviour: NGT Credentials were listed in plan as additions even after the operation was performed.

Reason: vms/uuid and vms/ API doesn't return NGT credentials as a part of nutanix guest customization. Due to which we were populating empty credentials in the state which caused the diff.


shreevari (approved):

venkatavivek-ntnx (approved):

venkatavivek-ntnx (approved):

yannickstruyf3 (approved):
LGTM

PR #310, opened 2021-11-25
NGT Creds: Added defensive check to update credentials when it is not passed in configuration

siddharth-nutanix:


nutanix/structure.go
```diff
@@ -374,8 +374,10 @@ func flattenNutanixGuestTools(d *schema.ResourceData, guest *v3.GuestToolsStatus
 		return err
 	}
 
-	if err := d.Set("ngt_credentials", ngtCredentials); err != nil {
-		return err
+	if _, ok := d.GetOk("ngt_credentials"); !ok {
+		if err := d.Set("ngt_credentials", ngtCredentials); err != nil {
+			return err
+		}
```
shreevari, line 380: Will this skip if we want to update ngt credentials on a VM which already has ngt installed with some creds? Is what I’m describing  a valid use case?

shreevari (approved):

nutanix/structure.go
siddharth-nutanix: Um, AFAIK we use ngt credentials just to configure a NGT on VM. Also, V3 APIs doesn't return ngt_credentials in the response.

venkatavivek-ntnx (approved):

PR #318, opened 2021-12-10
Update path to github.com/nutanix/terraform-provider-nutanix

embik:
This PR updates the current `github.com/terraform-providers/terraform-provider-nutanix` path in `go.mod` and include directives to `github.com/nutanix/terraform-provider-nutanix`, which is the new repository location. It seems that in the migration from `github.com/terraform-providers` (although the old path now redirects to `github.com/hashicorp/terraform-provider-nutanix`), this wasn't updated.

### DIFF
--- a/client/karbon/karbon_api.go
+++ b/client/karbon/karbon_api.go
@@ -1,7 +1,7 @@
 package karbon
 
 import (
-	"github.com/terraform-providers/terraform-provider-nutanix/client"
+	"github.com/nutanix/terraform-provider-nutanix/client"
 )
 
 const (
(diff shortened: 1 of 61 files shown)


basraayman (approved):

embik:
Actually, thanks to another Nutanix repo, I found a way to override this in a go.mod file (I've been trying various ways to do it, but didn't get to this): https://github.com/nutanix/docker-machine/blob/master/go.mod#L7. Possibly this shouldn't get merged as dependent projects would break? Not sure.

tuxtof:
there is plan to isolate nutanix sdk in a dedicated repo soon , so it will solve all this gimmick :-D

embik:
I'll go ahead and close this then, as the workaround from nutanix/docker-machine's `go.mod` works too.

PR #323, opened 2022-01-12
Add filter by cluster uuid in subnet datasource

### LINKED ISSUE #308: Allow Subnet Datasources to filter based on PE ID
The terraform provider currently (v1.2.x) fails when the subnet data source is used to fetch a subnet with a non-unique name.

We should provide the capability to (optionally) filter subnets based on PE ID

shreevari:
Fixes #308


yannickstruyf3 (changes requested):
Can we create a generic "filter" param so it is easy to extend it with additional parameters down the road? For example filter subnets based on VLAN ID?
This same filter logic can be applied to other data sources as well. Similar example: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/ami

shreevari:
AWS provider uses a common ec2 filter struct and [transparently passes them to the SDK](https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/ec2/common_schema_data_source.go#L17-L20), thus being handled by SDK/API.

An equivalent of this in Nutanix V3 APIs is the filter param in the metadata, which we already expose for list type data sources. 
Do you suggest that I add a filter param for single data sources like `nutanix_subnet`, `nutanix_cluster`, etc and pass the filter through to the API call?

shreevari:
Here's an example of filtering for list type data sources that I tried out for filtering by `vlan_id`
```

data "nutanix_clusters" "clusters" {}

output "sbnet" {
  value = data.nutanix_subnets.test
}

data "nutanix_subnets" "test" {
  metadata {
    filter = "vlan_id==154"
  }
}
```

siddharth-nutanix:
@shreevari Can you also add an example and update the datasource documentation for this change?

client/client.go
```diff
@@ -239,6 +246,138 @@ func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) error
...
+
+	return err
+}
+
+func filter(body io.ReadCloser, filters []*AdditionalFilter, baseSearchPaths [][]string) io.ReadCloser {
+	if filters == nil {
+		return body
+	}
+	var res map[string]interface{}
+	b, _ := io.ReadAll(body)
```
siddharth-nutanix: Do not ignore the error.

client/client.go
```diff
@@ -239,6 +246,138 @@ func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) error
...
+			searchPath := append(baseSearchPath, strings.Split(filter.Name, ".")...)
+			filterSearchPaths = append(filterSearchPaths, searchPath)
+		}
+		searchPaths[filter.Name] = filterSearchPaths
+	}
+
+	// Entities that pass filters
+	var filteredEntities []interface{}
+
+	entities := res["entities"].([]interface{})
```
siddharth-nutanix, line 335: Why have we hardcoded "entities" here ? Will all Prism list APIs have "entities" as the top level json key in the response ?

client/client.go
```diff
@@ -239,6 +246,138 @@ func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) error
...
+	for _, entity := range entities {
```
siddharth-nutanix, line 336: Can we bring down the complexity of this loop ? 
We can also think of using external library like [jsonpath](https://github.com/PaesslerAG/jsonpath)

client/v3/v3_service.go
```diff
@@ -57,7 +57,7 @@ type Service interface {
 	DeleteVolumeGroup(uuid string) error
 	CreateVolumeGroup(request *VolumeGroupInput) (*VolumeGroupResponse, error)
 	ListAllVM(filter string) (*VMListIntentResponse, error)
-	ListAllSubnet(filter string) (*SubnetListIntentResponse, error)
+	ListAllSubnet(filter string, additionalFilters []*client.AdditionalFilter) (*SubnetListIntentResponse, error)
```
siddharth-nutanix: filter --> all filters available on API
additionalFilters --> client side filters? 

Can we rename additionalFilters to clientSideFilters for clarity?

siddharth-nutanix (changes requested):

client/client.go
shreevari: Yes, all Prism list APIs have `entities` as top level json key in response

siddharth-nutanix:
/ok-to-test
>Line code coverage is 60.2

yannickstruyf3 (approved):

client/client.go
```diff
@@ -239,6 +246,136 @@ func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) error
...
+	if err != nil {
+		return err
+	}
+
+	if v != nil {
+		if w, ok := v.(io.Writer); ok {
+			_, err = io.Copy(w, resp.Body)
+			if err != nil {
+				fmt.Printf("Error io.Copy %s", err)
+
```
yannickstruyf3: Can we remove unnecessary newlines?

client/client.go
```diff
@@ -239,6 +246,136 @@ func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) error
...
+// DoWithFilters performs request passed and filters entities in json response
+func (c *Client) DoWithFilters(ctx context.Context, req *http.Request, v interface{}, filters []*AdditionalFilter, baseSearchPaths []string) error {
+	req = req.WithContext(ctx)
+	resp, err := c.client.Do(req)
+
+	if err != nil {
+		return err
+	}
+
+	defer func() {
```
yannickstruyf3, line 265: Indentation of the code and spacings are not consistent with rest of the code

client/client.go
```diff
@@ -239,6 +246,136 @@ func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) error
...
+	var filteredEntities []interface{}
+
+	entities := res["entities"].([]interface{})
+	for _, entity := range entities {
+		filtersPassed := 0
+	filter_loop:
+		for filter, filterSearchPaths := range searchPaths {
+			for _, searchPath := range filterSearchPaths {
+				searchTarget := entity.(map[string]interface{})
+
```
yannickstruyf3: remove newline

siddharth-kulshrestha (approved):

siddharth-nutanix (approved):

yannickstruyf3 (changes requested):
Can we also add a test case for filters in data_source_nutanix_subnet_test.go?

client/client.go
shreevari: I've run `gofmt` to iron out indentation issues. Is the spacing inconsistent due to unnecessary newlines? I've removed a bunch of newlines just now.

yannickstruyf3 (approved):

PR #368, opened 2022-03-02
Data source for discovering nodes in foundation

### LINKED ISSUE #357: [Foundation]: Add datasource for discovery of nodes
Add datasource for discovery of nodes

Use /foundation/discover_nodes api to create a datasource for discovery of nodes

Main Feature : #354

bhatipradeep:


abhimutant (approved):

client/foundation/foundation_api.go
```diff
@@ -45,6 +48,9 @@ func NewFoundationAPIClient(credentials client.Credentials) (*Client, error) {
 		FileManagement: FileManagementOperations{
 			client: client,
 		},
+		Networking: NetworkingOperations{
+			client: client,
```
siddharth-kulshrestha, line 52: Make sure that client is not writing to any global variable. As we are using a single client object in all foundation service's clients it may cause race condition in parallel execution or cause unwanted behaviour.

nutanix/data_source_foundation_discover_nodes.go
```diff
@@ -0,0 +1,168 @@
...
+		node["node_uuid"] = v.NodeUUID
+		node["current_network_interface"] = v.CurrentNetworkInterface
+		node["node_position"] = v.NodePosition
+		node["hypervisor"] = v.Hypervisor
+		node["configured"] = v.Configured
+		node["nos_version"] = v.NosVersion
+
+		// to make sure we set accept integer response
+		if v.ClusterID != nil && reflect.TypeOf(v.ClusterID).String() != "string" {
+			node["cluster_id"] = int64((v.ClusterID).(float64))
```
siddharth-kulshrestha: can we directly convert to int64? 
Also, this is a risky conversion that can also return error.

nutanix/data_source_foundation_discover_nodes.go
```diff
@@ -0,0 +1,168 @@
...
+		node["ipv6_address"] = v.Ipv6Address
+		node["node_uuid"] = v.NodeUUID
+		node["current_network_interface"] = v.CurrentNetworkInterface
+		node["node_position"] = v.NodePosition
+		node["hypervisor"] = v.Hypervisor
+		node["configured"] = v.Configured
+		node["nos_version"] = v.NosVersion
+
+		// to make sure we set accept integer response
+		if v.ClusterID != nil && reflect.TypeOf(v.ClusterID).String() != "string" {
```
siddharth-kulshrestha: reflect package generally make code execution slow. If we can assert it directly and check for error it will be a better option.

nutanix/data_source_foundation_discover_nodes.go
```diff
@@ -0,0 +1,168 @@
...
+			node["cluster_id"] = int64((v.ClusterID).(float64))
+		} else {
+			node["cluster_id"] = int64(-1)
+		}
+
+		// to make sure we set accept integer response
+		if v.CurrentCvmVlanTag != nil && reflect.TypeOf(v.CurrentCvmVlanTag).String() != "string" {
```
siddharth-kulshrestha: Same as above two comments.

siddharth-kulshrestha (approved):

client/foundation/foundation_api.go
bhatipradeep: Yes @siddharth-kulshrestha . Parallel datasource/resource running can cause issues ?

bhatipradeep: If yes we need to do something about it

nutanix/data_source_foundation_discover_nodes.go
bhatipradeep: According to docs for interface{} type, the underlying type after unmarshalling is float64, so it needed to be asserted with float64 then it can be typecasted to int64

siddharth-nutanix: [Check this.
](https://stackoverflow.com/questions/18041334/convert-interface-to-int)

We should also handle error scenario where this conversion is not possible.

siddharth-nutanix (approved):

PR #379, opened 2022-03-07
Task/resource api keys

abhimutant:


client/fc/fc_service.go
```diff
@@ -22,9 +22,9 @@ type Service interface {
 	CreateCluster(input *CreateClusterInput) (*CreateClusterResponse, error)
 	UpdateCluster(clusterUUID string, updateData *UpdateClusterData) error
 	DeleteCluster(clusterUUID string) error
-	// CreateAPIKey()
```
siddharth-kulshrestha, line 25: Remove these commented lines.

client/fc/fc_service.go
```diff
@@ -22,9 +22,9 @@ type Service interface {
...
-	// GetAPIKey()
-	// ListAPIKeys()
+	CreateAPIKey(input *CreateAPIKeysInput) (*CreateAPIKeysResponse, error)
```
siddharth-kulshrestha: We should pass context (ctx) in all these functions.

nutanix/data_source_nutanix_foundation_central_api_keys.go
```diff
@@ -0,0 +1,64 @@
...
+		if err := d.Set("current_time", resp.CurrentTime); err != nil {
+			return diag.FromErr(err)
+		}
+		if err := d.Set("alias", resp.Alias); err != nil {
+			return diag.FromErr(err)
+		}
+
+		d.SetId(uuid.(string))
+	} else {
+		return diag.Errorf("please provide `uuid`")
```
siddharth-kulshrestha:
```suggestion
		return diag.Errorf("please provide `key_uuid`")
```

nutanix/data_source_nutanix_foundation_central_list_api_keys.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+}
+
+func dataSourceNutanixFCListAPIKeysRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	conn := meta.(*Client).FC
+
+	req := &fc.ListMetadataInput{}
+	length, lok := d.GetOk("length")
+	if lok {
+		req.Length = utils.IntPtr(length.(int))
```
siddharth-kulshrestha, line 84: If length and offset are not given then do we get all the response entities ? 
I guess we don't get there will be one upper limit in case of other APIs we are calling the same API again and again until we get all the entities.

nutanix/data_source_nutanix_foundation_central_list_api_keys.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+	offset, ok := d.GetOk("offset")
+	if ok {
+		req.Offset = utils.IntPtr(offset.(int))
+	}
+
+	resp, err := conn.ListAPIKeys(req)
+	if err != nil {
+		return diag.Errorf("error reading API keys with error %s", err)
```
siddharth-kulshrestha: err is of type error, I dont think so that it can be printed here.

siddharth-kulshrestha (approved):

nutanix/data_source_nutanix_foundation_central_list_api_keys.go
abhimutant: yes. we get all responses even we don't give length and offset params.

PR #389, opened 2022-03-17
Make PC connection details optional with foundation connection details

bhatipradeep:
Changes :
- Make PC connection details and Foundation details optional in provider configuration.
- Update client creation code to return empty client in case required details are not present.
- Raise warnings about disabled provider services as per the provider configuration fields
- Raise errors if certain resource is used without giving proper provider configuration fields required.
- Fix existing v3 tests to have some dummy endpoint value rather then "" because now there will be a case of empty client affected due to it as well.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -176,6 +176,11 @@ func NewBaseClient(credentials *Credentials, absolutePath string, isHTTP bool) (
 
 // NewRequest creates a request
 func (c *Client) NewRequest(ctx context.Context, method, urlStr string, body interface{}) (*http.Request, error) {
+	// check if client exists or not
+	if c.client == nil {
+		return nil, fmt.Errorf("client for %s is missing", c.UserAgent)
+	}
+
 	rel, errp := url.Parse(c.AbsolutePath + urlStr)
 	if errp != nil {
 		return nil, errp
@@ -214,6 +219,11 @@ func (c *Client) NewRequest(ctx context.Context, method, urlStr string, body int
 
 // NewRequest creates a request without authorisation headers
 func (c *Client) NewUnAuthRequest(ctx context.Context, method, urlStr string, body interface{}) (*http.Request, error) {
+	// check if client exists or not
+	if c.client == nil {
+		return nil, fmt.Errorf("client for %s is missing", c.UserAgent)
+	}
+
 	//create main api url
 	rel, err := url.Parse(c.AbsolutePath + urlStr)
 	if err != nil {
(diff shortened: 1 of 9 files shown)


client/client.go
```diff
@@ -176,6 +176,11 @@ func NewBaseClient(credentials *Credentials, absolutePath string, isHTTP bool) (
 
 // NewRequest creates a request
 func (c *Client) NewRequest(ctx context.Context, method, urlStr string, body interface{}) (*http.Request, error) {
+	// check if client exists or not
+	if c.client == nil {
+		return nil, fmt.Errorf("client for %s is missing", c.UserAgent)
```
siddharth-kulshrestha: Improve the error message. User should get an idea on what parameters he is missing.

siddharth-kulshrestha (approved):

client/client.go
```diff
@@ -49,6 +49,9 @@ type Client struct {
 
 	// absolutePath: for example api/nutanix/v3
 	AbsolutePath string
+
+	// error message, incase client is in error state
+	Error string
```
siddharth-kulshrestha:
```suggestion
	ErrorMsg string
```

client/v3/v3.go
```diff
@@ -18,16 +18,25 @@ type Client struct {
...
+
+	// check if all required fields are present. Else create an empty client
+	if credentials.Username != "" && credentials.Password != "" && credentials.Endpoint != "" {
+		c, err := client.NewClient(&credentials, userAgent, absolutePath, false)
+		if err != nil {
+			return nil, err
+		}
+		baseClient = c
+	} else {
+		baseClient = &client.Client{UserAgent: userAgent, Error: "nutanix pc Client is missing. " +
```
siddharth-kulshrestha:
```suggestion
		baseClient = &client.Client{UserAgent: userAgent, Error: "nutanix PC (Prism Central) client is missing. " +
```

siddharth-kulshrestha: Pass these required parameters from ConfigureProvider function. Otherwise in future we need to make changes in this package as well if we change auth schema for any client. 

Make changes in every client.

Even we can think of creating auth fields as parameters of credentials. We can create error msg dynamically once we know the keys or fields of auth schema.

Make changes for all clients.

nutanix/provider.go
```diff
@@ -160,15 +165,46 @@ func Provider() *schema.Provider {
...
 }
 
 // This function used to fetch the configuration params given to our provider which
 // we will use to initialize a dummy client that interacts with API.
-func providerConfigure(d *schema.ResourceData) (interface{}, error) {
+func providerConfigure(ctx context.Context, d *schema.ResourceData) (interface{}, diag.Diagnostics) {
 	log.Printf("[DEBUG] config wait_timeout %d", d.Get("wait_timeout").(int))
 
+	// required fields for individual provider service
+	requiredProviderFields := map[string][]string{
```
siddharth-kulshrestha: Can we derieve these from constants ?

siddharth-kulshrestha (changes requested):

siddharth-kulshrestha (approved):

PR #432, opened 2022-04-27
Optimize Image upload to avoid buffering. Add cluster related fields in image upload resource & data source for PC

### LINKED ISSUE #428: Not able to upload image to a specific cluster
Problem Statement:
Image can be uploaded to a specific cluster incase of multi-cluster PC setup. Currently nutanix_image resource doesn't allow any attribute to perform such operation.

### LINKED ISSUE #430: Image upload failures - Help me help you stop buffering your POSTs
Foundation 5.1.1
This would also apply to any cluster or PC version as it has nothing to do with any of these
### Terraform Version
[wwhitt@phx-se-automation-dev-1 test]$ terraform -v
+ Terraform v1.1.9 on linux_amd64 (also tested on v1.1.2)
+ provider registry.terraform.io/nutanix/nutanix v1.5.0-beta.2
### Affected Resource(s)

* nutanix_foundation_image 
*  foundation_file_management_service 
*  client/client.go 
*  v3_service.go
* ### Terraform Configuration Files

```hcl
[wwhitt@phx-se-automation-dev-1 test]$ cat main.tf
terraform {
  required_providers {
    nutanix = {
      source = "nutanix/nutanix"
      version = "1.5.0-beta.2"
    }
  }
}

provider "nutanix" {
  foundation_endpoint= "10.55.76.52"
}

// upload aos image
resource "nutanix_foundation_image" "image1" {
  source = "/home/wwhitt/tmp/AOS/6.1/nutanix_installer_package-release-fraser-6.1-stable-d0d01a21610008feb6f566fbfe4aa767bda26e0a-x86_64.tar.gz"
  filename = "nos-tempfile-6.1.tar"
  installer_type = "nos"
}

data "nutanix_foundation_nos_packages" "nos" {
  depends_on = [resource.nutanix_foundation_image.image1]
}

output "nos" {
  value = data.nutanix_foundation_nos_packages.nos
}

```

### Panic Output
```Stack trace from the terraform-provider-nutanix_v1.5.0-beta.2 plugin:

fatal error: runtime: out of memory

runtime stack:
runtime.throw({0xd30f11, 0x118c00000})
	runtime/panic.go:1198 +0x71
runtime.sysMap(0xc23a000000, 0x4298e0, 0xc00006be90)
	runtime/mem_linux.go:169 +0x96
runtime.(*mheap).grow(0x13aa460, 0x8c479)
	runtime/mheap.go:1393 +0x225
runtime.(*mheap).allocSpan(0x13aa460, 0x8c479, 0x0, 0x1)
	runtime/mheap.go:1179 +0x165
runtime.(*mheap).alloc.func1()
	runtime/mheap.go:913 +0x69
runtime.systemstack()
	runtime/asm_amd64.s:383 +0x49

goroutine 39 [running]:
runtime.systemstack_switch()
	runtime/asm_amd64.s:350 fp=0xc00054f2c8 sp=0xc00054f2c0 pc=0x461420
runtime.(*mheap).alloc(0x1188f2000, 0x8c479, 0xd3, 0x0)
	runtime/mheap.go:907 +0x73 fp=0xc00054f318 sp=0xc00054f2c8 pc=0x425c13
runtime.(*mcache).allocLarge(0xc00054f3f8, 0x1188f2000, 0x60, 0x1)
	runtime/mcache.go:227 +0x89 fp=0xc00054f378 sp=0xc00054f318 pc=0x416889
runtime.mallocgc(0x1188f2000, 0x0, 0x0)
	runtime/malloc.go:1088 +0x5c5 fp=0xc00054f3f8 sp=0xc00054f378 pc=0x40ce05
runtime.growslice(0xc0002f85d8, {0xc090356000, 0x2ce3c000, 0x2ce3c000}, 0xb38ec001)
	runtime/slice.go:261 +0x4ac fp=0xc00054f460 sp=0xc00054f3f8 pc=0x44a56c
io.ReadAll({0xe33e60, 0xc0002f85d8})
	io/io.go:631 +0xa5 fp=0xc00054f4e0 sp=0xc00054f460 pc=0x48c3a5
io/ioutil.ReadAll(...)
	io/ioutil/ioutil.go:27
```
### Expected Behavior
It uploads an image without chewing through at least (imageSize * 2) worth of memory. Chewing up more than 1G is probably unreasonable regardless of incoming filesize. Imagine uploading a Windows qcow 😄  while it eats windows qcow * 2 memory doing it.
### Actual Behavior
SIGSEGV; Fails to upload because we read the entire file into memory, then we clone that into a new buffer.... THEN we pass it to net/http. double buffering. mmmmmm
### Steps to Reproduce
1. `terraform apply` where image size > mem+swap allocated to machine 
### Important Factors
*os.File implements io.reader which is what net/http is looking for, so if the only reason we're reading the entire file to spawn a new buffer is to satisfy the net/http NewRequest format... that's not necessary
Here's an example where I do the same thing against the same Foundation API endpoint, but I'm simply using base go libraries and passing the file ptr to http.Post:
```[wwhitt@phx-se-automation-dev-1 test]$ go run test.go
{
  "md5sum": null, 
  "name": "/home/nutanix/foundation/nos/go-upload-2-aos-6.1.tar.gz", 
  "in_whitelist": true
}
[wwhitt@phx-se-automation-dev-1 test]$ cat test.go 
package main

import (
	"io"
	"fmt"
	"os"
	"net/http"
)

func main() {
	file, err := os.Open("/home/wwhitt/tmp/AOS/6.1/nutanix_installer_package-release-fraser-6.1-stable-d0d01a21610008feb6f566fbfe4aa767bda26e0a-x86_64.tar.gz")
	if err != nil { fmt.Print(err) }
	defer file.Close()

	resp, err := http.Post("http://10.55.76.52:8000/foundation/upload?installer_type=nos&filename=go-upload-2-aos-6.1.tar.gz", "application/octet-stream", file)
	defer resp.Body.Close()
	ret, err := io.ReadAll(resp.Body)
	fmt.Print(string(ret))
} 
```
^ the above code is stupid simple, but I've profiled both and actually streaming as shown in the example above never went above like .4% memory usage, compared to balloon -> balloon -> balloon -> crash. and as you can see, the file uploaded successfully.

bhatipradeep:
New additions: 
1. Directly used os.File as input to http.NewRequest() and added implementation for setting content length & GetBody values for req{} as there is no implementation in http.NewRequest() for os.File type .
2. Added cluster related fields in image upload resource and data source for PC to upload image to particular cluster in multicluster setup

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -309,6 +309,10 @@ func (c *Client) NewUploadRequest(ctx context.Context, method, urlStr string, fi
 		return nil, err
 	}
 	req.ContentLength = fileInfo.Size()
+	req.GetBody = func() (io.ReadCloser, error) {
+		r := *fileReader
+		return io.NopCloser(&r), nil
+	}
 
 	req.Header.Add("Content-Type", octetStreamType)
 	req.Header.Add("Accept", mediaType)
@@ -346,6 +350,10 @@ func (c *Client) NewUnAuthUploadRequest(ctx context.Context, method, urlStr stri
 		return nil, err
 	}
 	req.ContentLength = fileInfo.Size()
+	req.GetBody = func() (io.ReadCloser, error) {
+		r := *fileReader
+		return io.NopCloser(&r), nil
+	}
 
 	req.Header.Add("Content-Type", octetStreamType)
 	req.Header.Add("Accept", mediaType)
--- a/client/client.go
+++ b/client/client.go
@@ -310,7 +310,7 @@ func (c *Client) NewUploadRequest(ctx context.Context, method, urlStr string, fi
 }
 
 // NewUploadRequest handles image uploads for image service
-func (c *Client) NewUnAuthUploadRequest(ctx context.Context, method, urlStr string, body []byte) (*http.Request, error) {
+func (c *Client) NewUnAuthUploadRequest(ctx context.Context, method, urlStr string, file *os.File) (*http.Request, error) {
 	// check if client exists or not
 	if c.client == nil {
 		return nil, fmt.Errorf(c.ErrorMsg)
@@ -322,9 +322,7 @@ func (c *Client) NewUnAuthUploadRequest(ctx context.Context, method, urlStr stri
 
 	u := c.BaseURL.ResolveReference(rel)
 
-	buf := bytes.NewBuffer(body)
-
-	req, err := http.NewRequest(method, u.String(), buf)
+	req, err := http.NewRequest(method, u.String(), file)
 
 	if err != nil {
 		return nil, err
--- a/client/client.go
+++ b/client/client.go
@@ -301,8 +301,8 @@ func (c *Client) NewUploadRequest(ctx context.Context, method, urlStr string, fi
 		return nil, err
 	}
 
-	// Set req.ContentLength by reading file size, as we send *os.File in http.NewRequest()
-	// http.NewRequest() only sets content length value for types - bytes.Buffer, bytes.Reader and strings.Reader
+	// Set req.ContentLength and req.GetBody as internally there is no implementation of such for os.File type reader
+	// http.NewRequest() only sets this values for types - bytes.Buffer, bytes.Reader and strings.Reader
 	// Refer https://github.com/golang/go/blob/a0f77e56b7a7ecb92dca3e2afdd56ee773c2cb07/src/net/http/request.go#L896
 	fileInfo, err := fileReader.Stat()
 	if err != nil {
@@ -310,8 +310,7 @@ func (c *Client) NewUploadRequest(ctx context.Context, method, urlStr string, fi
 	}
 	req.ContentLength = fileInfo.Size()
 	req.GetBody = func() (io.ReadCloser, error) {
-		r := *fileReader
-		return io.NopCloser(&r), nil
+		return io.NopCloser(fileReader), nil
 	}
 
 	req.Header.Add("Content-Type", octetStreamType)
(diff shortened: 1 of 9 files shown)


client/client.go
```diff
@@ -323,16 +335,26 @@ func (c *Client) NewUnAuthUploadRequest(ctx context.Context, method, urlStr stri
...
+	req, err := http.NewRequest(method, u.String(), fileReader)
 
-	req, err := http.NewRequest(method, u.String(), buf)
+	if err != nil {
+		return nil, err
+	}
 
+	// Set req.ContentLength and req.GetBody as internally there is no implementation of such for os.File type reader
+	// http.NewRequest() only sets this values for types - bytes.Buffer, bytes.Reader and strings.Reader
+	// Refer https://github.com/golang/go/blob/a0f77e56b7a7ecb92dca3e2afdd56ee773c2cb07/src/net/http/request.go#L896
```
WWhitt, line 346: ~Nit: While this is true for those data types, it doesn't seem necessary when using an *os.File, see my example in the issue~ -- it uploaded perfectly fine without pre-emptively specifying content-length. That being said, I totally support not relying on default behavior here.

Edit: while I did get uploads without doing that, it wasn't working how I thought. ignore me. I agreed with explicitly setting it anyway

siddharth-kulshrestha (approved):

siddharth-nutanix:
/ok-to-test
>Acceptance test run status: failure
>

bhati-pradeep:
Testcase failed due to setup config mismatch. Passing locally after fixing. Merging it to master. 

Attached the screenshot.

Screenshot 2022-04-29 at 1 18 25 PM

PR #468, opened 2022-06-21
Feat/m static routes

Abhishekism9450:


nutanix/provider.go
```diff
@@ -162,6 +162,7 @@ func Provider() *schema.Provider {
 			"nutanix_foundation_central_imaged_node_details":  dataSourceFCImagedNodeDetails(),
 			"nutanix_vpc":                                     dataSourceNutanixVPC(),
 			"nutanix_vpc_list":                                dataSourceNutanixVPCList(),
+			"nutanix_static_routes":                           dataSourceNutanixStaticRoute(),
```
siddharth-kulshrestha, line 169: Do we need a singular datasource for this?

siddharth-kulshrestha: Is it possible ?

nutanix/resource_nutanix_static_routes.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+
+	if _, errWaitTask := stateConf.WaitForStateContext(ctx); errWaitTask != nil {
+		return diag.Errorf("error waiting for vpc Static Route (%s) to create: %s", d.Id(), errWaitTask)
+	}
+
+	return resourceNutanixStaticRouteRead(ctx, d, meta)
+}
+
+func resourceNutanixStaticRouteDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
siddharth-kulshrestha, line 217: We need to discuss this. As you mentioned that we can delete the static route from UI.

nutanix/provider.go
siddharth-kulshrestha: Also for plural datasources we should stick to one convention either : nutanix_datasource_types or nutanix_datasource_type_list for new datasources.

siddharth-kulshrestha (changes requested):

siddharth-kulshrestha (approved):

PR #477, opened 2022-07-19
Feat/m user groups

Abhishekism9450:


client/v3/v3_structs.go
```diff
@@ -2255,20 +2254,33 @@ type UserGroupSpec struct {
...
+}
+
 // User Group Resource Definition
 type UserGroupResources struct {
 	AccessControlPolicyReferenceList []*Reference               `json:"access_control_policy_reference_list,omitempty"` // List of ACP references.
 	DirectoryServiceUserGroup        *DirectoryServiceUserGroup `json:"directory_service_user_group,omitempty"`         // A Directory Service user group.
 	DisplayName                      *string                    `json:"display_name,omitempty"`                         // The display name for the user group.
 	ProjectsReferenceList            []*Reference               `json:"projects_reference_list,omitempty"`              // A list of projects the user group is part of.
 	UserGroupType                    *string                    `json:"user_group_type,omitempty"`
+	DirectoryServiceOU               *DirectoryServiceUserGroup `json:"directory_service_ou,omitempty"`
```
siddharth-kulshrestha, line 2269: Is the attribute name correct ?

nutanix/resource_nutanix_user_groups.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+
+	if err = d.Set("saml_user_group", flattenSamlUserGroup(resp.Spec.Resources.SamlUserGroup)); err != nil {
+		return diag.Errorf("error setting saml_user_group for user group %s: %s", d.Id(), err)
+	}
+
+	return nil
+}
+
+func resourceNutanixUserGroupsUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
siddharth-kulshrestha, line 198: Can't we update the existing user group from UI?
If not, can we make all attributes as ForceNew?

nutanix/resource_nutanix_user_groups.go
```diff
@@ -0,0 +1,275 @@
...
+}
+
+func resourceNutanixUserGroupsRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	conn := meta.(*Client).API
+
+	resp, err := conn.V3.GetUserGroup(d.Id())
+	if err != nil {
+		if strings.Contains(fmt.Sprint(err), "ENTITY_NOT_FOUND") {
+			d.SetId("")
+			return nil
```
bhati-pradeep: If entity is not found, we can return the error right ? Instead of nil.

nutanix/resource_nutanix_user_groups.go
```diff
@@ -0,0 +1,275 @@
...
+}
+
+func resourceNutanixUserGroupsDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	conn := meta.(*Client).API
+	log.Printf("[DEBUG] Deleting User Group: %s", d.Id())
+	resp, err := conn.V3.DeleteUserGroup(ctx, d.Id())
+	if err != nil {
+		if strings.Contains(fmt.Sprint(err), "ENTITY_NOT_FOUND") {
+			d.SetId("")
+			return nil
```
bhati-pradeep: same here

siddharth-kulshrestha (changes requested):

nutanix/resource_nutanix_user_groups.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+	d.SetId("")
+
+	return nil
+}
+
+func expandDirectoryUserGroup(pr []interface{}) *v3.DirectoryServiceUserGroup {
+	if len(pr) > 0 {
+		res := &v3.DirectoryServiceUserGroup{}
+		ent := pr[0].(map[string]interface{})
```
bhati-pradeep, line 314: Why just [0] ?

client/v3/v3_structs.go
Abhishekism9450: yes. OU means organisational unit.

nutanix/resource_nutanix_user_groups.go
Abhishekism9450: There is no UI for user group. Also, API doesn't support to update user group as well.

nutanix/resource_nutanix_user_groups.go
Abhishekism9450: there will be only one object in list.

nutanix/resource_nutanix_user_groups.go
Abhishekism9450: Done.

nutanix/resource_nutanix_user_groups.go
Abhishekism9450: Done

PR #490, opened 2022-09-07
Feat/m era sla datasource

Abhishekism9450:
DataSource for Era :
1. SLA (by id and name)
2. SLAs (list)
3. Cluster (by id and name)
4. Clusters (list)
5. Profile
6. Profiles


client/era/era_service.go
```diff
@@ -0,0 +1,158 @@
...
+
+	return res, sc.c.Do(ctx, httpReq, res)
+}
+
+func (sc ServiceClient) GetSLA(ctx context.Context, id string, name string) (*ListSLAResponse, error) {
+	var path string
+	if id != "" {
+		path = fmt.Sprintf("/slas/%s", id)
+	}
+	if name != "" {
```
bhati-pradeep, line 86: else if would be better here

client/era/era_service.go
```diff
@@ -0,0 +1,158 @@
...
+		}
+		return path
+	}
+
+	if name != "" {
+		path := "/profiles?name=" + name
+		return path
+	}
+	return ""
+}
```
bhati-pradeep, line 158: In this file code at multiple places there are continuous multiple "if" clauses. Can we use "else-if" to avoid edge cases to be at safer side

client/era/era_structs.go
```diff
@@ -0,0 +1,139 @@
...
+
+type ProfileListResponse []ListProfileResponse
+
+type Propertiesmap struct {
+	DefaultContainer *string `json:"DEFAULT_CONTAINER"`
+	MaxVdiskSize     *string `json:"MAX_VDISK_SIZE"`
+}
+
+type Properties struct {
+	RefID       interface{} `json:"ref_id,omitempty"`
```
bhati-pradeep: is the return type of RefID not constant ? thats why we used interface here ?

client/era/era_structs.go
```diff
@@ -0,0 +1,139 @@
...
+	Ownerid                *string `json:"ownerId,omitempty"`
+	Datecreated            *string `json:"dateCreated,omitempty"`
+	Datemodified           *string `json:"dateModified,omitempty"`
+	CurrentActiveFrequency *string `json:"currentActiveFrequency,omitempty"`
+	Continuousretention    int     `json:"continuousRetention,omitempty"`
+	Dailyretention         int     `json:"dailyRetention,omitempty"`
+	Weeklyretention        int     `json:"weeklyRetention,omitempty"`
+	Monthlyretention       int     `json:"monthlyRetention,omitempty"`
+	Quarterlyretention     int     `json:"quarterlyRetention,omitempty"`
+	Yearlyretention        int     `json:"yearlyRetention,omitempty"`
```
bhati-pradeep, line 133: Please check if value 0 matters here or not. If yes, then int should be *int

bhati-pradeep: Check other places as well

nutanix/data_source_nutanix_era_profiles.go
```diff
@@ -0,0 +1,424 @@
...
+			ercs["optimized_for_provisioning"] = v.OptimizedForProvisioning
+
+			res[k] = ercs
+		}
+		return res
+	}
+	return nil
+}
+
+func flattenPropertiesMap(prm map[string]interface{}) map[string]interface{} {
```
bhati-pradeep: Do we need this seperate function ? as its just a small check

nutanix/provider.go
```diff
@@ -111,6 +111,12 @@ func Provider() *schema.Provider {
 				DefaultFunc: schema.EnvDefaultFunc("FOUNDATION_PORT", nil),
 				Description: descriptions["foundation_port"],
 			},
+			"era_endpoint": {
```
siddharth-kulshrestha, line 117: Does ERA supports basic auth or some kind of authentication? 
What if PC auth is different from ERA auth and user wants to use both PC APIs and ERA APIs?

nutanix/data_source_nutanix_era_cluster_test.go
```diff
@@ -0,0 +1,70 @@
+package nutanix
+
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+)
+
+func TestAccEraClusterDataSource_basic(t *testing.T) {
+	// r := randIntBetween(31, 40)
```
siddharth-kulshrestha, line 10:
```suggestion
```

nutanix/data_source_nutanix_era_clusters.go
```diff
@@ -0,0 +1,252 @@
...
+	}
+	d.SetId(uuid)
+	return nil
+}
+
+func flattenClustersResponse(crsp *Era.ClusterListResponse) []map[string]interface{} {
+	if crsp != nil {
+		lst := []map[string]interface{}{}
+		for _, v := range *crsp {
+			d := map[string]interface{}{}
```
siddharth-kulshrestha, line 189: Shouldn't we reuse the setter code of the singular datasource here ?

nutanix/data_source_nutanix_era_slas.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	d.SetId(uuid)
+	return nil
+}
+
+func flattenSLAsResponse(sla *Era.SLAResponse) []map[string]interface{} {
+	if sla != nil {
+		lst := []map[string]interface{}{}
+		for _, data := range *sla {
+			d := map[string]interface{}{}
```
siddharth-kulshrestha, line 121: Try to reuse or create common function for plural / singular datasources.

Make changes everywhere.

siddharth-kulshrestha (changes requested):

client/era/era_service.go
Abhishekism9450: each "if" statement has a return path.

client/era/era_structs.go
Abhishekism9450: changed it to "string" type .

client/era/era_structs.go
Abhishekism9450: these are for datasources.

nutanix/data_source_nutanix_era_clusters.go
Abhishekism9450: singular datasource are directly set w/o any method.

nutanix/data_source_nutanix_era_slas.go
Abhishekism9450: used common function for flattening but can't use setter of singular datasource as they are directly set on readContext .

nutanix/provider.go
Abhishekism9450: Era has basic auth. 
Era plugin has it's own creds ( username, password and endpoint ) .

nutanix/data_source_nutanix_era_profiles.go
Abhishekism9450: updated the code. Removed the func.

abhimutant:
/ok-to-test era
>Current line coverage (3.6) is below threshold (50). Kindly add more acceptance tests.

client/client.go
```diff
@@ -72,6 +72,9 @@ type Credentials struct {
 	FoundationEndpoint string              // Required field for connecting to foundation VM APIs
 	FoundationPort     string              // Port for connecting to foundation VM APIs
 	RequiredFields     map[string][]string // RequiredFields is client to its required fields mapping for validations and usage in every client
+	EraEndpoint        string              // Required field for connecting to Era VM APIs.
+	EraUsername        string
```
HiteshAmbarkhane, line 76: Era has been renamed to NDB.

HiteshAmbarkhane (commented):
Era has been rebranded as NDB. 
As this is public facing library. Unfortunately, this name change is needed everywhere.

Issue #494, opened 2022-09-20
Updating gives error: Machine type must be set to Q35 for secure boot.


ppanula:
### Nutanix Cluster Information
Nutanix Prism Element version:  5.20.4.6 LTS

### Terraform Version
Terraform v1.2.9
on linux_amd64
+ provider registry.terraform.io/hashicorp/template v2.2.0
+ provider registry.terraform.io/nutanix/nutanix v1.7.1

### Affected Resource(s)
* nutanix_virtual_machine

### Terraform Configuration Files
https://gist.github.com/ppanula/7d275086d6ba3492e92206122affbdcf

### Debug Output
https://gist.github.com/ppanula/5936d08df3126a97c3e4d247cef23ef6
### Expected Behavior
Updating memory from 2GB to 4 GB
### Actual Behavior
Error message: Machine type must be set to Q35 for secure boot.
### Steps to Reproduce

Make VM with SECURE_BOOT and Q35 then try to update memory

ppanula:
It seems actually that any terraform apply will fail for this same error message.

johnnyf-hcp:
I do see that @ppanula has submitted a PR for this in Sep 2022.
https://github.com/nutanix/terraform-provider-nutanix/pull/496

I have a customer who is hitting the same issue and it's currently blocking their POC.  Can somebody help to review this PR?

abhimutant:
Fix merged in master !

johnnyf-hcp:
Thanks @abhimutant, much appreciated!

PR #496, opened 2022-09-22
Fix Secure boot VMs when doing updates

ppanula:
This is my simple fix for SECURE_BOOT VMs to handle terraform apply updates, when it's doing first power off VM and then powering it back on. 

I noticed that MachineType was not set to Q35 when powering off was done and second time MachineType is missing when actual update is done and VM is booted after that.

Issue #494

### DIFF
--- a/nutanix/resource_nutanix_virtual_machine.go
+++ b/nutanix/resource_nutanix_virtual_machine.go
@@ -1337,6 +1337,12 @@ func changePowerState(ctx context.Context, conn *v3.Client, id string, powerStat
 	spec.AvailabilityZoneReference = response.Status.AvailabilityZoneReference
 	spec.ClusterReference = response.Status.ClusterReference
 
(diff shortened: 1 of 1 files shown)


nutanix/resource_nutanix_virtual_machine.go
```diff
@@ -1203,6 +1203,12 @@ func resourceNutanixVirtualMachineUpdate(ctx context.Context, d *schema.Resource
 		metadata.SpecVersion = specVersion
 	}
 
+	// Check if we are dealing with SECURE_BOOT machine, MachineType is needed then set to Q35
```
abhimutant, line 1206: we don't need this check at two places. we can always preFill the machineType using the response.
here: https://github.com/nutanix/terraform-provider-nutanix/blob/033e0f76532dc2c8df09befa355e35798b4db6d4/nutanix/resource_nutanix_virtual_machine.go#L928

abhimutant:
Closing as fix is already merged in master. #633

PR #509, opened 2022-11-06
Add documentation for ip_config_pool_list_ranges in subnet resource

Bermos:
The parameter was not yet documented, but is implemented and works.


website/docs/r/subnet.html.markdown
```diff
@@ -63,6 +63,7 @@ resource "nutanix_subnet" "next-iac-managed" {
 * `dhcp_server_address_port`: - (Optional) Port Number.
 * `dhcp_options`: - (Optional) Spec for defining DHCP options.
 * `dhcp_domain_search_list`: - (Optional).
+* `ip_config_pool_list_ranges`: - (Optional) Range of IPs (example: 10.0.0.9 10.0.0.19).
```
abhimutant, line 67: can you please provide correct examples for ip_config_pool_list_ranges. It should be like :
 ["10.0.0.9 10.0.0.19"] .

PR #513, opened 2022-11-09
filter based on database type

Abhishekism9450:
Support filter based on database type in GET call for databases.

### DIFF
--- a/nutanix/data_source_nutanix_ndb_databases.go
+++ b/nutanix/data_source_nutanix_ndb_databases.go
@@ -6,6 +6,7 @@ import (
 	"github.com/hashicorp/go-uuid"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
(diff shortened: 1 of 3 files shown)


nutanix/data_source_nutanix_ndb_databases.go
```diff
@@ -229,3 +247,14 @@ func flattenDatabaseIntancesList(db *era.ListDatabaseInstance) []map[string]inte
 	}
 	return nil
 }
+
+func filterDatabaseBasedOnDatabaseEngine(resp *era.ListDatabaseInstance, dbengine string) (*era.ListDatabaseInstance, error) {
+	found := make(era.ListDatabaseInstance, 0)
+
+	for _, v := range *resp {
+		if dbengine == v.Type {
```
siddharth-kulshrestha, line 255: @abhimutant Please convey this to ERA team to support this as a parameter.

siddharth-kulshrestha (approved):

nutanix/data_source_nutanix_ndb_databases.go
Abhishekism9450: It's already conveyed and we are tracking this.

PR #515, opened 2022-11-30
Feat/m ndb database restore

Abhishekism9450:


nutanix/resource_nutanix_ndb_database_restore.go
```diff
@@ -0,0 +1,406 @@
...
+
+func resourceNutanixNDBDatabaseRestoreRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	c := meta.(*Client).Era
+	if c == nil {
+		return diag.Errorf("era is nil")
+	}
+
+	databaseInstanceID := d.Id()
+
+	resp, err := c.Service.GetDatabaseInstance(ctx, databaseInstanceID)
```
siddharth-kulshrestha: Why are we reading database instance here?

siddharth-kulshrestha: Does the read tells us that on which snapshot ID or timestamp the database is presently at?

nutanix/resource_nutanix_ndb_database_restore.go
```diff
@@ -0,0 +1,406 @@
...
+		if err := d.Set("linked_databases", flattenDBLinkedDbs(resp.Linkeddatabases)); err != nil {
+			return diag.FromErr(err)
+		}
+	}
+
+	return nil
+}
+
+func resourceNutanixNDBDatabaseRestoreUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	return nil
```
siddharth-kulshrestha: Same problem here we need to have logic similar to create... because if user wants to restore at a particular point after making one restore he doesn't need to write another block.

siddharth-kulshrestha (changes requested):

nutanix/resource_nutanix_ndb_database_restore.go
bhati-pradeep: @siddharth-kulshrestha maybe the reason is that restore of database is a state change of database itself. @Abhishekism9450 Isnt this method implemented in datasource of database or in databases ? cant we directly use it here.

nutanix/resource_nutanix_ndb_database_restore.go
```diff
@@ -0,0 +1,472 @@
...
+		if err := d.Set("linked_databases", flattenDBLinkedDbs(resp.Linkeddatabases)); err != nil {
+			return diag.FromErr(err)
+		}
+	}
+
+	return nil
+}
+
+func resourceNutanixNDBDatabaseRestoreUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	conn := meta.(*Client).Era
```
bhati-pradeep: call resourceNutanixNDBDatabaseRestoreCreate() directly. The implementation is similar to it.

nutanix/resource_nutanix_ndb_log_catchups.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+	d.SetId(resp.Operationid)
+	return nil
+}
+
+func resourceNutanixNDBLogCatchUpsRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	return nil
+}
+
+func resourceNutanixNDBLogCatchUpsUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
bhati-pradeep, line 131: Call resourceNutanixNDBLogCatchUpsCreate() here as implementation is similar

bhati-pradeep (changes requested):

nutanix/resource_nutanix_ndb_database_restore.go
Abhishekism9450: Done.

nutanix/resource_nutanix_ndb_log_catchups.go
Abhishekism9450: Done.

nutanix/resource_nutanix_ndb_database_restore.go
Abhishekism9450: implemented restore_count to restore agian at same point of time or with same snapshot id.

nutanix/common_era_schema.go
```diff
@@ -1,12 +1,26 @@
...
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	era "github.com/terraform-providers/terraform-provider-nutanix/client/era"
 )
 
+type dbID string
+
+const dbIDKey dbID = ""
+
+func NewContext(ctx context.Context, dbID dbID) context.Context {
```
siddharth-kulshrestha, line 18: Can we also add a comment in the code explaining why is this done?

siddharth-kulshrestha (approved):

nutanix/common_era_schema.go
Abhishekism9450: sure.

PR #517, opened 2022-11-30
Feat/m ndb database scale

Abhishekism9450:
Resource for NDB: database scale


nutanix/resource_nutanix_ndb_database_scale.go
```diff
@@ -0,0 +1,421 @@
...
+		}
+
+		if err := d.Set("linked_databases", flattenDBLinkedDbs(resp.Linkeddatabases)); err != nil {
+			return diag.FromErr(err)
+		}
+	}
+	return nil
+}
+
+func resourceNutanixNDBScaleDatabaseUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
siddharth-kulshrestha, line 275: As discussed we need to add similar logic as of create to allow users to edit the same resource block and update the database size.

siddharth-kulshrestha (changes requested):

nutanix/resource_nutanix_ndb_database_scale.go
```diff
@@ -0,0 +1,499 @@
...
+			},
+			"pre_script_cmd": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"post_script_cmd": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"scale_count": {
```
bhati-pradeep, line 41: What does "scale_count" do ?

nutanix/resource_nutanix_ndb_database_scale.go
```diff
@@ -0,0 +1,499 @@
...
+				Type:     schema.TypeInt,
+				Optional: true,
+			},
+
+			// Computed values
+			"name": {
```
bhati-pradeep, line 47: The response will be database info right ? So can we import database instance datas source schema here ?

nutanix/resource_nutanix_ndb_database_scale.go
Abhishekism9450: scale_count helps to perform db scale with same config (data_storage_size) . Change the count and it will perform db scale ops with same config.

bhati-pradeep (approved):
LGTM

abhimutant (approved):

PR #529, opened 2022-12-20
Feat/m era profiles

Abhishekism9450:
Resources for profiles:-
1. Software
2. Network
3. Compute
4. Database Params

Added examples and testcases


nutanix/resource_nutanix_ndb_profiles.go
```diff
@@ -0,0 +1,1280 @@
...
+			},
+			"compute_profile": {
+				Type:          schema.TypeList,
+				Optional:      true,
+				ConflictsWith: []string{"software_profile", "network_profile", "database_parameter_profile"},
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"cpus": {
+							Type:     schema.TypeString,
+							Optional: true,
```
bhati-pradeep, line 48: Aren't this fields required ?

nutanix/resource_nutanix_ndb_profiles.go
```diff
@@ -0,0 +1,1280 @@
...
+			"network_profile": {
+				Type:          schema.TypeList,
+				Optional:      true,
+				ConflictsWith: []string{"compute_profile", "software_profile", "database_parameter_profile"},
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"topology": {
+							Type:         schema.TypeString,
+							Required:     true,
+							ValidateFunc: validation.StringInSlice([]string{"cluster", "single"}, false),
```
bhati-pradeep, line 123: Isn't "ALL" supported ?

nutanix/resource_nutanix_ndb_profiles.go
```diff
@@ -0,0 +1,1280 @@
...
+		req.Type = utils.StringPtr("Database_Parameter")
+		req.SystemProfile = false
+		req.DBVersion = utils.StringPtr("ALL")
+	}
+
+	if sp, ok := d.GetOk("software_profile"); ok {
+		softwareProfile = true
+		splist := sp.([]interface{})
+
+		for _, v := range splist {
```
bhati-pradeep, line 624: Why a loop if only one software profile create is only allowed ?

bhati-pradeep: Are we handling different versions here ?

nutanix/resource_nutanix_ndb_profiles.go
```diff
@@ -0,0 +1,1280 @@
...
+	}
+
+	// Now if published is present args
+
+	if publish, ok := d.GetOk("published"); ok {
+		req := &era.ProfileRequest{}
+		netReq := &era.UpdateProfileRequest{}
+
+		req.Published = publish.(bool)
+		res, err := conn.Service.GetProfiles(ctx, "", "", d.Id(), "")
```
bhati-pradeep: This function call is not readable.
Currently GetProfiles takes params: engine string, profileType string, id string and name string  as input.  We can create a 'filters' struct for this method having above params as attributes. Future changes will be easy.

bhati-pradeep: Something like GetProfile(filters *profileFilters)

nutanix/resource_nutanix_ndb_profiles.go
```diff
@@ -0,0 +1,1280 @@
...
+
+				for a := range st {
+					sublist[a] = utils.StringPtr(st[a].(string))
+				}
+				req.AvailableClusterIds = sublist
+			}
+		}
+	}
+
+	versionID := res.Versions[0].ID
```
bhati-pradeep, line 886: Versions[0] is not always the case of all kind of profiles right ? Software profile can have >1 profile versions

nutanix/resource_nutanix_ndb_profiles.go
```diff
@@ -0,0 +1,1280 @@
...
+
+			if hIns, ok := inst["ha_instance"]; ok && len(hIns.([]interface{})) > 0 {
+				prop = expandNetworkHAInstance(ctx, meta, hIns.([]interface{}))
+			}
+		}
+	}
+	return prop
+}
+
+func buildDatabaseProfileProperties(ps []interface{}) []*era.ProfileProperties {
```
bhati-pradeep, line 968: This method is technically for just postgres type database right ?

nutanix/resource_nutanix_ndb_profiles.go
```diff
@@ -0,0 +1,1280 @@
...
+	if len(ps) > 0 {
+		for _, v := range ps {
+			val := v.(map[string]interface{})
+			if psdb, ok := val["postgres_database"]; ok {
+				brr := psdb.([]interface{})
+
+				for _, v := range brr {
+					val := v.(map[string]interface{})
+
+					if p1, ok1 := val["max_connections"]; ok1 {
```
bhati-pradeep: There are around more than 10 properties to be set. Instead of setting for each props, we can set it in a loop as most of them have same structure and types. 
```suggestion
					for key, val := range brr:
					        prop = append(prop, {name:key, value:val, secure: False})
					
```

bhati-pradeep: Something like this ^

nutanix/resource_nutanix_ndb_profiles_test.go
```diff
@@ -0,0 +1,230 @@
...
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccEraPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccEraProfileConfigByCompute(name, desc),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceNameProfile, "name", name),
+					resource.TestCheckResourceAttr(resourceNameProfile, "description", desc),
+					resource.TestCheckResourceAttr(resourceNameProfile, "versions.#", "1"),
```
bhati-pradeep, line 24: We need to check value of cpus, memory, etc. as only 1 version is there.

nutanix/resource_nutanix_ndb_software_version_profile.go
```diff
@@ -0,0 +1,411 @@
...
+	stateConf := &resource.StateChangeConf{
+		Pending: []string{"PENDING"},
+		Target:  []string{"COMPLETED", "FAILED"},
+		Refresh: eraRefresh(ctx, conn, opReq),
+		Timeout: d.Timeout(schema.TimeoutCreate),
+		Delay:   eraDelay,
+	}
+
+	if _, errWaitTask := stateConf.WaitForStateContext(ctx); errWaitTask != nil {
+		return diag.Errorf("error waiting for software profile	version (%s) to create: %s", *resp.EntityID, errWaitTask)
```
bhati-pradeep: Remove extra spec before "version"

bhati-pradeep (changes requested):

nutanix/resource_nutanix_ndb_profiles.go
Abhishekism9450: No. If you check the method implementation. it will check for db_types and assign the properties values.

nutanix/resource_nutanix_ndb_profiles.go
Abhishekism9450: These are optional. If not given, defaults will be set .

nutanix/resource_nutanix_ndb_profiles.go
Abhishekism9450: schema is a list. Given max item 1 .

siddharth-kulshrestha (approved):

nutanix/resource_nutanix_ndb_profiles.go
Abhishekism9450: Implemented.

bhati-pradeep (approved):
LGTM

abhimutant (approved):

PR #530, opened 2023-01-12
added bad Request

Abhishekism9450:
Added conditional statement to check for bad request in client API .

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -550,7 +550,11 @@ func CheckResponse(r *http.Response) error {
 	// Nutanix returns non-json response with code 401 when
 	// invalid credentials are used
 	if c == http.StatusUnauthorized {
-		return fmt.Errorf("invalid Nutanix Credentials")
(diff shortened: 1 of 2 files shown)


client/client.go
```diff
@@ -550,7 +550,11 @@ func CheckResponse(r *http.Response) error {
 	// Nutanix returns non-json response with code 401 when
 	// invalid credentials are used
 	if c == http.StatusUnauthorized {
-		return fmt.Errorf("invalid Nutanix Credentials")
+		return fmt.Errorf("invalid Credentials")
```
siddharth-kulshrestha:
```suggestion
		return fmt.Errorf("invalid auth credentials")
```

siddharth-kulshrestha (approved):

client/client.go
Abhishekism9450: Done.

abhimutant (approved):

PR #541, opened 2023-01-30
refresh clone

Abhishekism9450:
Clone Refresh based on : 
1. snapshot id
2. timestamp

### DIFF
--- a/nutanix/resource_nutanix_ndb_clone_refresh.go
+++ b/nutanix/resource_nutanix_ndb_clone_refresh.go
@@ -1,0 +1,114 @@
+package nutanix
+
+import (
+	"context"
+	"log"
+	"time"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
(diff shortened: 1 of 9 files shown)


nutanix/resource_nutanix_ndb_clone_refresh.go
```diff
@@ -0,0 +1,114 @@
...
+
+	resp, err := conn.Service.RefreshClone(ctx, req, cloneID)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	// Get Operation ID from response of clone refresh and poll for the operation to get completed.
+	opID := resp.Operationid
+	if opID == "" {
+		return diag.Errorf("error: operation ID is an empty string")
```
bhati-pradeep, line 81: If there is no operation triggered then do we require this to error out ?

bhati-pradeep (approved):
LGTM

nutanix/resource_nutanix_ndb_clone_refresh.go
Abhishekism9450: Operation is triggered but response doesn't have operation id to poll. so , we throw error saying there is no operation id to poll .

abhimutant (approved):

PR #562, opened 2023-03-03
vlanid 0 is supported

Abhishekism9450:
Vlan ID with 0 is supported.

### DIFF
--- a/nutanix/resource_nutanix_subnet.go
+++ b/nutanix/resource_nutanix_subnet.go
@@ -717,7 +717,7 @@ func getSubnetResources(d *schema.ResourceData, subnet *v3.SubnetResources) {
 		dhcpo.DomainSearchList = expandStringList(v.([]interface{}))
 	}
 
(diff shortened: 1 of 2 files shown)


nutanix/resource_nutanix_subnet.go
```diff
@@ -717,7 +717,7 @@ func getSubnetResources(d *schema.ResourceData, subnet *v3.SubnetResources) {
 		dhcpo.DomainSearchList = expandStringList(v.([]interface{}))
 	}
 
-	if v, ok := d.GetOk("vlan_id"); ok {
+	if v, ok := d.GetOkExists("vlan_id"); ok {
 		if v.(int) == 0 || ok {
```
bhati-pradeep: I dont think we require this condition now, as ok would be always true

bhati-pradeep (approved):

abhimutant (approved):

PR #641, opened 2023-10-12
fix: compare bootconfig against previous value

### LINKED ISSUE #640: VM rebooted at every change because of `hotPlugChange` set to `false`
Hello,

We have noticed that some of our VMs get rebooted every time we run a `terraform apply`.
Those VMs have a custom boot configuration: they use the UEFI BIOS mode.

We suspect the following [piece of code](https://github.com/nutanix/terraform-provider-nutanix/blob/master/nutanix/resource_nutanix_virtual_machine.go#L1180) to be the cause of the reboots as it sets `hotPlugChange` to `false`:
```go
	if bc, change := bootConfigHasChange(res.BootConfig, d); !reflect.DeepEqual(*bc, v3.VMBootConfig{}) {
		res.BootConfig = bc
		hotPlugChange = change
	}
```

The following happens when it enters the `bootConfigHasChange` function:
* `hotPlugChange` is set to `false`;
* none of the conditions are matched, because there are no planned changes related to the boot config of the VM;
* the function returns the unchanged boot config and `false`.

Right after the `DeepEqual` check is executed and compares the boot config of the VM with an empty/default VM boot config. In our case the boot type is set to UEFI, so `DeepEqual` returns `false` and `hotPlugChange` is assigned the value of `change` - i.e. `false`.

I would be happy to prepare a PR to fix this (I guess `DeepEqual` should be comparing `bc` with `res.BootConfig` instead) but before that it would be great if somebody could review and confirm our assumptions.

FYI we run Terraform 1.5.5 with the latest version of the Nutanix provider (1.9.3).

Thanks,
Fred

fad3t:
fixes #640

### DIFF
--- a/nutanix/resource_nutanix_virtual_machine.go
+++ b/nutanix/resource_nutanix_virtual_machine.go
@@ -1251,19 +1251,20 @@ func getVMSpecVersion(conn *v3.Client, vmID string) (*int64, error) {
 func bootConfigHasChange(boot *v3.VMBootConfig, d *schema.ResourceData) (*v3.VMBootConfig, bool) {
 	hotPlugChange := false
 
-	if boot == nil {
-		boot = &v3.VMBootConfig{}
+	var bc = &v3.VMBootConfig{}
+	if boot != nil {
+		*bc = *boot
 	}
 
 	if d.HasChange("boot_device_order_list") {
 		_, n := d.GetChange("boot_device_order_list")
-		boot.BootDeviceOrderList = expandStringList(n.([]interface{}))
+		bc.BootDeviceOrderList = expandStringList(n.([]interface{}))
 		hotPlugChange = false
 	}
 
 	if d.HasChange("boot_type") {
 		_, n := d.GetChange("boot_type")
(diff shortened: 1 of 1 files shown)


nutanix/resource_nutanix_virtual_machine.go
```diff
@@ -1929,7 +1930,7 @@ func preFillResUpdateRequest(res *v3.VMResources, response *v3.VMIntentResponse)
 	if response.Spec.Resources.BootConfig != nil {
 		res.BootConfig = response.Spec.Resources.BootConfig
 	} else {
-		res.BootConfig = nil
+		res.BootConfig = &v3.VMBootConfig{}
 	}
```
abhimutant: we cannot pass null object for BootConfig spec as API restrict us. This will throw error incase user hasn't specified BootConfig in tf file.

abhimutant:
LGTM

abhimutant (approved):

bhati-pradeep (approved):
LGTM

PR #644, opened 2023-10-20
changing the param to be Optional

Abhishekism9450:
external subnet name/uuid are Optional args not Required .


bhati-pradeep (approved):
LGTM

PR #703, opened 2024-12-23
Removed the modules used deprecated v4 sdks

Haroon-Dweikat-Ntx:
endpoints which are deprecated needs to be removed:

  1. List all the iSCSI attachments associated with the given Volume Group
  2. List all the VM attachments for a Volume Group
  3. List all the category details that are associated with the Volume Group


abhimutant (approved):
LGTM

PR #776, opened 2025-03-18
Install nodejs before terraform setup

GullapalliAkhil:
The hashicorp/setup-terraform action is built using JavaScript (Node.js). When GitHub Actions runs this step, it expects node to be installed to execute the Terraform setup script. If node is missing, the action cannot execute, causing the below error.
/usr/bin/env: ‘node’: No such file or directory
##[error***Process completed with exit code 127.

To solve the above error made changes to install nodejs as well before terraform setup


bhati-pradeep:
LGTM

PR #784, opened 2025-03-20
Add support for showing skipped testcases in test summary

GullapalliAkhil:
As of now in test summary we doesn't show the Skipped testcases count and the reason for skip. With this changes we have added the support for showing skipped testcases in summary.


GullapalliAkhil (commented):
LGTM!!

PR #786, opened 2025-03-20
skip pc restore tests based on skip_pc_restore_test variable

Haroon-Dweikat-Ntx:
## TESTCASES 

Skip the `TestAccV2NutanixRestorePCResource` test cases based on a variable defined in` test_confog_v2`. We are skipping the PC restore tests because they require powering off the PC VM, which will affect the execution of other test cases running in parallel and it will fail. The PC restore test cases can be run separately.


GullapalliAkhil (approved):
LGTM

PR #787, opened 2025-03-21
fix backup target failed tests

Haroon-Dweikat-Ntx:
## TESTCASE: 
- fixed prismv2 backup target failed tests due to taking the first backup target, now we taking it based on cluster ext id 
- add more tests to cover both cluster location and object store location [backup target and restore source]


GullapalliAkhil (approved):
LGTM!

GullapalliAkhil (approved):

Issue #852, opened 2025-05-26
[Feat] Develop terraform resource for ovaimages


kvandenhoute:
**Is your feature request related to a problem? Please describe.**
Right now, it is possible to create a vm from OVA image in Prism. It is however not possible to do this with the terraform provider. Therefor, the only way to deploy a vm from image is to have a raw image. However, if you want to download + upload a third party OVA and deploy that, you should first deploy a vm from the OVA and then create an image from it and only then can you deploy it using terraform. This manual step to create a VM inbetween should be removed.

**Describe the solution you'd like**
Have a data resource OVAImage with which you can read OVAImage properties. This can then be used to create a new vm. The OVAImage returns basically a VMSpec, so all these properties can then be used in nutanix_virtual_machine resource. Not in the nutanix_virtual_machine_v2 resource, as the v4 API does not support data_source_reference from url (which is needed to deploy from OVA). The v3 API does support this. 

**Describe alternatives you've considered**
Adapt the nutanix_virtual_machine(_v2) spec to have a property ova_image, but working this way it would be harder to adapt / extend properties

**Additional context**

PR #853, opened 2025-05-26
[Feat] Develop terraform resource for nutanix_ova_image

kvandenhoute:
Implements #852 

The usage works as following:
```hcl
data "nutanix_ova_image" "image" {
  ova_image_name = "myova.img"
}

resource "nutanix_virtual_machine" "vm" {
  // vm specific properties
  name                 = var.hostname
  cluster_uuid = data.nutanix_cluster.cluster[var.hostname].id
  num_vcpus_per_socket = var.num_cpu
  num_sockets          = var.num_cores_per_socket
  memory_size_mib      = var.memory
  
  // re-use properties from OVA image
  boot_device_order_list = data.nutanix_ova_image.image.boot_device_order_list
  boot_type = data.nutanix_ova_image.image.boot_type

  // The most important bit, is to use the disks  
  dynamic "disk_list" {
    for_each = { 
      for i, template in data.nutanix_ova_image.image.disk_list: 
      i => template if template.device_properties[0].device_type == "DISK"
    }
    content {
      data_source_reference = {
        url = disk_list.value.data_source_reference.url
      }
      device_properties {
        device_type = disk_list.value.device_properties[0].device_type
        disk_address = disk_list.value.device_properties[0].disk_address
      }
      disk_size_bytes = disk_list.value.disk_size_bytes
    }
  }
  
  // Subnets can also be reused
  dynamic "nic_list" {
    for_each = var.networks
    content {
      subnet_uuid =  ${nic_list.value[0]}
    }
  }
}
```

This however, does not sit 100% right with terraform, as when the VM is read after creation, the disk_list.*.data_source_reference does not have the initial URL but has a UUID instead as the disk is created (url is not returned by the api). Meaning, every apply would mean a change.

Therefore the nutanix_ova_image usage requires to ignore disk_list, and to allow for updates to come through, to use a null_resource and replace_triggered_by.

```hcl
resource "nutanix_virtual_machine" "vm" {
  ...
  lifecycle {
    replace_triggered_by = [ null_resource.image_trigger ]
    ignore_changes = [ 
      disk_list
     ]
  }
  depends_on = [null_resource.image_trigger]
}

locals {
  image_replace_hash = sha1(jsonencode({
    ova_image_name = data.nutanix_ova_image.image.ova_image_name
    disk_urls      = [for d in data.nutanix_ova_image.image.disk_list : d.data_source_reference.url]
  }))
}

resource "null_resource" "image_trigger" {
  triggers = {
    image_hash = local.image_replace_hash
  }
}
```

If any changes are required, please let me know. I think this solution is one of the only feasible ones, within the limitations of the existing nutanix API (v3 / v4)


GullapalliAkhil:
@kvandenhoute This is more of a v3 implementation?

kvandenhoute:
Hi @GullapalliAkhil 

Yes, as the V4 API does not seem the have an endpoint for /OVAs, nor does the V4 API of virtual machines have an option to use a url as datasource for disks.

Please correct me if I'm wrong, as I'd rather use the latest and greatest. But I did not find it in the API docs and noticed prism uses V3 for this functionality too

GullapalliAkhil:
@kvandenhoute Currently V4 API does have an endpoint for OVAs but it is not GA'ed yet. In next releases after TF v2.3.0(which is not yet released) we are planning to develop this functionality using V4.

PR #860, opened 2025-05-29
Fix issue 857: How to use boot_config under nutanix_virtual_machine_v2

### LINKED ISSUE #857: How to use boot_config under nutanix_virtual_machine_v2
Below is the description of boot_config given on the provider portal

[boot_config](https://registry.terraform.io/providers/nutanix/nutanix/latest/docs/resources/virtual_machine_v2#boot_config-1): (Optional) Indicates the order of device types in which the VM should try to boot from. If the boot device order is not provided the system will decide an appropriate boot device order.

This does not describe how to specify the boot order when using UEFI VM with boot_config. We will need the proper description as to how specify the boot order in UEFI mode.

Whenever, we create UEFI VM, we cannot use legacy boot options like boot_device and boot_device_order_list

In VM deployment automation, we should be able to specify the boot order with Disk, CDROM (in the same order). This configuration will ignore the disk and boot through CD during initial VM deployment through Terraform, and post install, VM should be able to boot through the disk and ignore the CDROM. Without this control, VM automation is going through an endless loop as it boot through the CDROM everytime causing endless loop installation of OS on the VM, requiring timely manual intervention to break the loop, thereby, taking away the essence of automation

Haroon-Dweikat-Ntx:
Update virtual machine examples and docs example with boot config details and how to use boot config

### DIFF
--- a/website/docs/r/virtual_machine_v2.html.markdown
+++ b/website/docs/r/virtual_machine_v2.html.markdown
@@ -46,40 +46,102 @@ resource "nutanix_virtual_machine_v2" "vm-2"{
             }
         }
     }
+    boot_config {
+        uefi_boot {
+            boot_order = ["NETWORK", "DISK", "CDROM", ]
+        }
+    }
 }
 
-resource "nutanix_virtual_machine_v2" "vm-3"{
-    name= "example-vm-3"
-    description =  "vm desc"
-    num_cores_per_socket = 1
-    num_sockets = 1
-    cluster {
-        ext_id = "1cefd0f5-6d38-4c9b-a07c-bdd2db004224"
+resource "nutanix_virtual_machine_v2" "vm-3" {
+  name                 = "terraform-example-vm-4-disks"
+  num_cores_per_socket = 1
+  num_sockets          = 1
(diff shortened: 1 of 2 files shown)


website/docs/r/virtual_machine_v2.html.markdown
```diff
@@ -46,40 +46,102 @@ resource "nutanix_virtual_machine_v2" "vm-2"{
...
+      nic_type = "NORMAL_NIC"
+      subnet {
+        ext_id = "7f66e20f-67f4-473f-96bb-c4fcfd487f16"
+      }
+      vlan_mode = "ACCESS"
+    }
+  }
+
+  boot_config {
+    legacy_boot {
```
GullapalliAkhil, line 138: Better add a config in doc string which conveys on how one can configure with UEFI boot as well.

GullapalliAkhil (approved):

website/docs/r/virtual_machine_v2.html.markdown
Haroon-Dweikat-Ntx: already added

PR #889, opened 2025-07-08
Setting placeholder credentials to align with security guidelines

jonas-werner:
Changing usernames and passwords included in sample ENV files to align with security guidelines, i.e. "<username>" and "<password>" rather than default credentials.


abhinavbansal29 (approved):
L.G.T.M.

PR #936, opened 2025-09-11
Add support for Cisco Intersight managed nodes onboarding and imaging

cgascoig:
This PR is intended to enable Terraform automation of the day 0 provisioning of Cisco UCS nodes managed by Intersight. 

The PR:
- Adds a new resource `nutanix_foundation_central_onboard_nodes` to onboard nodes in Foundation Central from a connected Intersight hardware manager
- Extends the `nutanix_foundation_central_image_cluster` resource with the `.server_configuration_data`, `.fc_api_key_uuid` and `.node_list[].server_configuration_data` attributes required for imaging clusters on Intersight managed nodes

### DIFF
--- a/nutanix/services/foundationCentral/resource_nutanix_foundation_central_onboard_nodes.go
+++ b/nutanix/services/foundationCentral/resource_nutanix_foundation_central_onboard_nodes.go
@@ -105,7 +105,7 @@ func resourceNutanixFCOnboardNodesCreate(ctx context.Context, d *schema.Resource
 		for _, node := range hwManagerNodes.Nodes {
 			if *node.NodeSerial == serial {
 				var req fc.CreateOnboardNodeInput
-				err := copier.Copy(&req, &node) //nolint:scopelint
+				err := copier.Copy(&req, node)
 				if err != nil {
 					return diag.FromErr(err)
 				}
--- a/nutanix/services/foundationCentral/resource_nutanix_foundation_central_onboard_nodes.go
+++ b/nutanix/services/foundationCentral/resource_nutanix_foundation_central_onboard_nodes.go
@@ -26,8 +26,6 @@ func ResourceNutanixFCOnboardNodes() *schema.Resource {
 		Schema: map[string]*schema.Schema{
 			"node_serial": {
 				Type:     schema.TypeString,
-				Computed: false,
-				Optional: false,
 				Required: true,
 			},
 			"block_serial": {
@@ -85,7 +83,6 @@ func resourceNutanixFCOnboardNodesRead(ctx context.Context, d *schema.ResourceDa
 		return diag.FromErr(err)
 	}
 
-	d.SetId(*resp.ImagedNodeUUID)
 	return nil
 }
 
--- a/nutanix/services/foundationCentral/resource_nutanix_foundation_central_onboard_nodes.go
+++ b/nutanix/services/foundationCentral/resource_nutanix_foundation_central_onboard_nodes.go
@@ -1,0 +1,152 @@
+package fc
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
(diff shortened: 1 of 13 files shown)


nutanix/services/foundationCentral/resource_nutanix_foundation_central_onboard_nodes.go
```diff
@@ -0,0 +1,152 @@
+package fc
+
+import (
+	"context"
```
GullapalliAkhil, line 4: Can we please add tests which helps us to execute all the scenarios around this?

cgascoig: I have now added an acceptance test in nutanix/services/foundationCentral/resource_nutanix_foundation_central_onboard_nodes_test.go

I've assumed that the test environment does not have an Intersight account integrated, so the test expects the onboarding to fail with node node found.

nutanix/services/foundationCentral/resource_nutanix_foundation_central_onboard_nodes.go
```diff
@@ -0,0 +1,152 @@
...
+		ReadContext:   resourceNutanixFCOnboardNodesRead,
+		UpdateContext: resourceNutanixFCOnboardNodesUpdate,
+		DeleteContext: resourceNutanixFCOnboardNodesDelete,
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(ImageMiniTimeout),
+		},
+		Schema: map[string]*schema.Schema{
+			"node_serial": {
+				Type:     schema.TypeString,
+				Computed: false,
```
Haroon-Dweikat-Ntx: You are explicitly setting Computed = false and Optional = false in the schema definition.
By default, both Computed and Optional are false. When the attribute is Required, you don’t need to redundantly specify them. This makes the schema cleaner and easier to read.

```

// Suggested (cleaner)
"node_serial": {
    Type:     schema.TypeString,
    Required: true,
}
```

nutanix/services/foundationCentral/resource_nutanix_foundation_central_onboard_nodes.go
```diff
@@ -0,0 +1,152 @@
...
+		return diag.FromErr(err)
+	}
+	if err := d.Set("node_state", resp.NodeState); err != nil {
+		return diag.FromErr(err)
+	}
+	if err := d.Set("node_type", resp.NodeType); err != nil {
+		return diag.FromErr(err)
+	}
+
+	d.SetId(*resp.ImagedNodeUUID)
```
Haroon-Dweikat-Ntx: in a Terraform Read function, you don’t normally need to call `d.SetId()` every time.
Terraform already knows the resource ID from when it was created. the read step is supposed to refresh the attributes, not re-set the ID. Calling `d.SetId()` again doesn’t break anything, but it’s redundant unless the API returns a different identifier than the one Terraform is tracking.

nutanix/services/foundationCentral/resource_nutanix_foundation_central_onboard_nodes_test.go
```diff
@@ -0,0 +1,38 @@
...
+				ExpectError: regexp.MustCompile(`Node not found with serial ABC12345D6E`),
+			},
+		},
+	})
+}
+
+func testFCOnboardNodesResource() string {
+	return `
+		resource "nutanix_foundation_central_onboard_nodes" "node" {
+			node_serial = "ABC12345D6E"
```
Haroon-Dweikat-Ntx: Instead of hard-coding a dummy `node_serial` directly in the configuration block, define it as a test variable in `main_test.go`. This keeps the test setup flexible and avoids embedding static values in the HCL config.

Haroon-Dweikat-Ntx (changes requested):

Haroon-Dweikat-Ntx (approved):
LGTM

GullapalliAkhil (approved):

GullapalliAkhil:
@cgascoig Can we add the dcoumentation, it helps!!

GullapalliAkhil:
LGTM!!

Issue #947, opened 2025-09-23
Bug Report: resource "nutanix_user_groups_v2"


DriftwoodEcho92:
## Nutanix Cluster Information
Prism Central Version: pc.7.3.0.5
AOS Version: 7.3.0.5

## Terraform Version
Terraform: v1.13.1
Provider: registry.terraform.io/nutanix/nutanix v2.3.1

## Affected Resources
- resource "nutanix_user_groups_v2"

## Terraform Configuration Files
```terraform
terraform {
  required_providers {
    nutanix = {
      source  = "nutanix/nutanix"
      version = "2.3.1"
    }
    random = {
      source ="hashicorp/random"
      version = "~> 3.0"
    }
  }
}

# Directory service vars

variable "directory_service_name" {
    type=string
}

variable "directory_service_type" {
    type=string
}

variable "endpoint" {
    type=string
}

variable "port" {
    type = string
}

variable "service_account_username" {
    type =string
}

variable "service_account_password" {
    type = string
}

variable "domain" {
    type = string
}

resource "nutanix_directory_services_v2" "directory-service" {
  name           = var.directory_service_name
  url            = "ldap://${var.endpoint}:${var.port}"
  directory_type = var.directory_service_type
  domain_name    = var.domain
  service_account {
    username = var.service_account_username 
    password = var.service_account_password 
  }
}

# User group vars

variable "group_distinguished_name" {
  type = string
}

variable "group_name" {
  type = string
}


resource "nutanix_user_groups_v2" "user_group"{
  group_type         = "LDAP"
  idp_id             = nutanix_directory_services_v2.directory-service.ext_id
  name               = var.group_name
  distinguished_name = var.group_distinguished_name
}
```

## Debug Output
```bash
```

## Expected Behaviour
- To have persistent attribute mapping for ext_id like the other resources.
- Support all CRUDI operations

## Actual Behaviour
- If I apply a configuration and the resources get's created and verification is done, and then run apply again terraform evaluates that a change is made but it's not...
- Inconsistant name of resource, should be "nutanix_user_group_v2" without an **s**.

#### Create
⚠️ 
Only **.id** attribute is filled, **.ext_id** is **null** on creation.
Creation is only possible if the user is found in the directory, else it fails with a raw api response

#### Read
⚠️
Does not read **ext_id** attribute

#### Update
❌
Missing update operation in v4 API. And not implemented in provider resoruce.

#### Delete
✔️
Seems okay

#### Import
❌
[...]

PR #964, opened 2025-10-04
Fix ext_id null issue for resource nutanix_user_groups_v2

GullapalliAkhil:
When a resource for nutanix_user_groups_v2 is created, in state file we are not storing the ext_id as we are storing the uuid of resource in id. As part of ticket https://github.com/nutanix/terraform-provider-nutanix/issues/947 we are adding this changes.


Haroon-Dweikat-Ntx (approved):
L.G.T.M

Issue #951, opened 2025-09-23
Bug Report: "nutanix_roles_v2"


DriftwoodEcho92:
## Nutanix Cluster Information
Prism Central Version: pc.7.3.0.5
AOS Version: 7.3.0.5

## Terraform Version
Terraform: v1.13.1
Provider: registry.terraform.io/nutanix/nutanix v2.3.1

## Affected Resources
- "nutanix_roles_v2"

## Terraform Configuration Files
```terraform
terraform {
  required_providers {
    nutanix = {
      source  = "nutanix/nutanix"
      version = "2.3.1"
    }
  }
}

# filtered list operation
data "nutanix_operations_v2" "operations-filtered-list" {
  filter = "startswith(displayName, 'Create_')"
}

# Create role
resource "nutanix_roles_v2" "example-role"{
  display_name = "example_role"
  description  = "some description"
  client_name = "something" // Does nothing
  operations = [
    data.nutanix_operations_v2.operations-filtered-list.operations[0].ext_id,
    data.nutanix_operations_v2.operations-filtered-list.operations[1].ext_id,
    data.nutanix_operations_v2.operations-filtered-list.operations[2].ext_id,
    data.nutanix_operations_v2.operations-filtered-list.operations[25].ext_id
  ]
}
```

## Expected Behaviour
- Proper description and validation for arguments and attributes.
- Support for import
- Proper naming convention, name "nutanix_role_v2"

## Actual Behaviour
- The attribute "ext_id" is set to null on create, read and update, **id** is set but becomes a discrepancy in using the provider.
- The argument `client_name` needs a proper description and validation, as of now it does not seem to do anything. On apply terraform accepts the value whatever it is.
- The name is "nutanix_roles_v2" which is not following a proper naming convention.

#### Create
⚠️ 
**client_name** argument needs proper description and validation.

#### Read
✔️
Seems okay

#### Update
✔️
Seems okay

#### Delete
✔️ 
Seems okay

#### Import
❌
Does not support import

## Steps To Reproduce
1. Create main.tf and define a resource
2. Issue `terraform apply` - **Create** is tested
3. Modify a computed attribute on the resurce via the api or the webui and issue `terraform refresh` - **Read** is tested
4. Modify an attribute in main.tf and run `terraform apply` - **Update** is tested
5. Issue terraform destroy - **Delete** is tested
[...]

PR #965, opened 2025-10-06
add ext_id to be shown in state file for a resource

GullapalliAkhil:
Currently when we create a role we are not setting the ext_id in state file, this PR fixes the discrepancy. As part of ticket https://github.com/nutanix/terraform-provider-nutanix/issues/951 we have added this change.


Haroon-Dweikat-Ntx (approved):
LGTM

PR #1051, opened 2026-01-05
Fix v4 client initialization (port parsing + disable version negotiation) 

Haroon-Dweikat-Ntx:
## Summary
This PR improves Prism v4 client initialization and makes the Password Manager v2 password-change flow more reliable.

## Changes
- **v4 SDK client init**
  - Default **port to 9440** when `credentials.Port` is empty/invalid (avoid setting port `0` or returning errors)
  - Set `AllowVersionNegotiation = false` on v4 clients that support it

- **Password Manager v2**
  - Add tolerant detection for “unauthorized” errors during task polling after a password change
  - Recreate the Prism client with the **new password** and retry task fetch (preserving “insecure” behavior)
  - Ensure `AllowVersionNegotiation` is disabled on the recreated Prism client

### DIFF
--- a/nutanix/sdks/v4/clusters/clusters.go
+++ b/nutanix/sdks/v4/clusters/clusters.go
@@ -6,6 +6,7 @@ import (
 	"github.com/nutanix/ntnx-api-golang-clients/clustermgmt-go-client/v4/api"
 	cluster "github.com/nutanix/ntnx-api-golang-clients/clustermgmt-go-client/v4/client"
 	"github.com/terraform-providers/terraform-provider-nutanix/nutanix/client"
+	"github.com/terraform-providers/terraform-provider-nutanix/nutanix/sdks/v4/sdkconfig"
 )
 
 type Client struct {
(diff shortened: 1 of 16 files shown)


nutanix/sdks/v4/clusters/clusters.go
```diff
@@ -27,13 +26,14 @@ func NewClustersClient(credentials client.Credentials) (*Client, error) {
...
+		pcClient.Port = 9440
+		if credentials.Port != "" {
+			if p, err := strconv.Atoi(credentials.Port); err == nil {
+				pcClient.Port = p
+			}
 		}
-		pcClient.Port = port
 		pcClient.VerifySSL = false
-
+		pcClient.AllowVersionNegotiation = false
```
abhinavbansal29: Can we store `false` in some constants file because whenever we need to enable it, then we can do by just changing at one place.

abhinavbansal29 (changes requested):

nutanix/sdks/v4/clusters/clusters.go
Haroon-Dweikat-Ntx: done,

GullapalliAkhil (approved):

abhinavbansal29 (approved):

Issue #961, opened 2025-09-30
disk create issue for nutanix_virtual_machine_v2


chkp-eviatars:
Similer to https://github.com/nutanix/terraform-provider-nutanix/issues/882

### Nutanix Cluster Information
Please provide the version of:
 - Nutanix Cluster 7.3
 - Nutanix Prism Central 7.3
### Terraform Version
Terraform v1.10.5
on windows_386
+ provider registry.terraform.io/nutanix/nutanix v2.3.0
### Affected Resource(s)

* nutanix_virtual_machine_v2
### Terraform Configuration Files

```hcl
resource "nutanix_virtual_machine_v2" "vm" {
.
.
.
  memory_size_bytes = var.mgmt_memory != "8589934592" ? var.mgmt_memory : "8589934592"  # Default 8GB in bytes
  cluster {
    ext_id = data.nutanix_cluster.cluster.id
  }
.
.
.

  disks {
    disk_address {
      bus_type = "SCSI"
      index    = 0
    }
    backing_info {
      vm_disk {
        disk_size_bytes = "107374182400" # 100GB - adjust as needed
        data_source {
          reference {
            image_reference {
              image_ext_id = data.nutanix_image.mgmt_image.id
            }
          }
        }
      }
    }
  }
.
.
.
  power_state = "ON"
  depends_on = [nutanix_vpc_v2.vpc_example]
  }
}
```
### Debug Output

 Error: Attribute must be a whole number, got 1.073741824e+11
│ 
│   with module.Demo-VPC.nutanix_virtual_machine_v2.mgmt_vm,
│   on modules\tenant-vpc\main.tf line 136, in resource "nutanix_virtual_machine_v2" "mgmt_vm":
│  136:       vm_disk {

### Expected Behavior
VM should created with 8GB ram and 100GB disk
### Actual Behavior
Creation did not finish
### Steps to Reproduce
create VPC (not required)
create  VM with 8 GB ram and 100 GB disk (QCOW image)
1. `terraform apply`
### Important Factors
* When set 1GB it works
* When comment the RAM, it workis with default 1GB ram.
* not working when set as string or int.
* looks like the number should be TypeInt64 instead TypeInt here: https://github.com/nutanix/terraform-provider-nutanix/blob/bb7f855526761924688a9f02eaa4d4aeacc0fdda/nutanix/services/vmmv2/resource_nutanix_virtual_machine_v2.go#L395

LarryWisherMan:
[...]

PR #1061, opened 2026-01-15
[Imprv] Add human-friendly GiB sizing for memory and disks in nutanix_virtual_machine_v2

### LINKED ISSUE #1060: [Imprv] Add human-friendly GiB sizing fields to nutanix_virtual_machine_v2 resource
## Description
Add alternative human-friendly sizing fields (`memory_size_gib`, `disk_size_gib`) to the `nutanix_virtual_machine_v2` resource and data source, similar to how AWS provider handles sizing.

## Motivation
Currently, all size fields require bytes (e.g., `memory_size_bytes = 4294967296`), which is not user-friendly. This improvement adds GiB alternatives that are easier to read and write.

## Changes
- Add `memory_size_gib` field (conflicts with `memory_size_bytes`)
- Add `disk_size_gib` fields for disks, nvram, and cd_rom
- Automatically compute GiB values when reading from API
- Maintain full backward compatibility with existing `*_bytes` fields

## Example Usage
```hcl
resource "nutanix_virtual_machine_v2" "vm" {
  memory_size_gib = 4   # Instead of memory_size_bytes = 4294967296
  disks {
    backing_info {
      vm_disk {
        disk_size_gib = 100  # Instead of disk_size_bytes = 107374182400
      }
    }
  }
}

### PR DESCRIPTION
**Pull request by Svaillan**
Adds memory_size_gib and disk_size_gib fields to nutanix_virtual_machine_v2 resource and data source. Closes #1060

### DIFF
--- a/nutanix/services/vmmv2/data_source_nutanix_virtual_machine_v2.go
+++ b/nutanix/services/vmmv2/data_source_nutanix_virtual_machine_v2.go
@@ -74,6 +74,10 @@ func DatasourceNutanixVirtualMachineV4() *schema.Resource {
 				Type:     schema.TypeInt,
 				Computed: true,
 			},
+			"memory_size_gib": {
+				Type:     schema.TypeInt,
+				Computed: true,
+			},
 			"is_vcpu_hard_pinning_enabled": {
 				Type:     schema.TypeBool,
 				Computed: true,
@@ -473,6 +477,10 @@ func DatasourceNutanixVirtualMachineV4() *schema.Resource {
 																Computed: true,
 															},
 															"disk_size_bytes": {
+																Type:     schema.TypeInt,
+																Computed: true,
+															},
+															"disk_size_gib": {
 																Type:     schema.TypeInt,
 																Computed: true,
 															},
@@ -1359,6 +1367,11 @@ func DatasourceNutanixVirtualMachineV4Read(ctx context.Context, d *schema.Resour
 	if err := d.Set("memory_size_bytes", getResp.MemorySizeBytes); err != nil {
 		return diag.FromErr(err)
 	}
+	if getResp.MemorySizeBytes != nil {
+		if err := d.Set("memory_size_gib", *getResp.MemorySizeBytes/(1024*1024*1024)); err != nil {
+			return diag.FromErr(err)
+		}
+	}
 	if err := d.Set("is_vcpu_hard_pinning_enabled", getResp.IsVcpuHardPinningEnabled); err != nil {
 		return diag.FromErr(err)
 	}
@@ -2034,6 +2047,7 @@ func flattenVMDisk(pr *config.VmDisk) []map[string]interface{} {
 
 		if pr.DiskSizeBytes != nil {
 			disks["disk_size_bytes"] = pr.DiskSizeBytes
+			disks["disk_size_gib"] = *pr.DiskSizeBytes / (1024 * 1024 * 1024)
 		}
 		if pr.StorageContainer != nil {
 			disks["storage_container"] = flattenVMDiskContainerReference(pr.StorageContainer)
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by chkp-eviatars** [comment]
Hey
Is that will support both Terraform 32bit and Terraform 64bit binary? in https://github.com/nutanix/terraform-provider-nutanix/issues/961 I encounter integer overflow in the 32Bit

**Comment by Svaillan** [comment]
> Hey Is that will support both Terraform 32bit and Terraform 64bit binary? in #961 I encounter integer overflow in the 32Bit

The `memory_size_gib` and `disk_size_gib` values should work on 32-bit and 64-bit Terraform. The GiB values fit in a 32-bit int, and I have set it up to cast to int64 before multiplying by 1024³.

This would probably provide a sort of workaround for #961 because you could use use memory_size_gib = 8 instead of memory_size_bytes = 8589934592. Anything over 2GB (2,147,483,647 bytes) will be over the overflow limit using 32 bit.

**Comment by chkp-eviatars** [comment]
Hi @Svaillan, Thank you for the answer.
Just to understand, have you test `disk_size_gib` and `memory_size_gib` with, lets say, 8gib mem and 100gb disk, in Win **32bit** and 64bit terraform binary?

**Comment by rutwij-nerkar** [comment]
Hi @Svaillan ,

Thanks for working on this. I pulled the branch, rebased it onto current `master` (2.4.2) and tested against a live cluster. A few things need rework before this is functionally complete:

1. `disk_size_gib` is never applied (create/update). The field is added to the schema and the read/flatten path, but neither `expandVMDisk` nor `expandOneOfDiskBackingInfo` reads it. So `disk_size_gib = 100` passes plan/validate (since `disk_size_bytes` is computed) but is silently ignored on apply — the disk is created without the requested size. Please wire `disk_size_gib → DiskSizeBytes` in both expand functions, mirroring the memory logic (cast to `int64` before `*1024^3`).

2. `ConflictsWith` breaks provider InternalValidate. `nutanix_ova_v2` (resource + data sources, added in 2.4.2) embeds the full VM schema via `ResourceNutanixVirtualMachineV2()`. `ConflictsWith` paths resolve from the resource root, so once embedded the bare `memory_size_gib`/`memory_size_bytes` references don't resolve and the provider fails to load:

```
resource nutanix_ova_v2: ConflictsWith: memory_size_bytes references unknown attribute (memory_size_gib)
Please drop ConflictsWith from these fields and enforce mutual exclusivity in code (you already give GiB precedence), or validate via CustomizeDiff.
```

3. Data-source schema is inconsistent for `disks. disk_size_gib` is only declared on the NVRAM `backing_storage_info` block, but the flatten functions write it for the main `disks[].backing_info.vm_disk` and `cd_roms[].backing_info` blocks too. Please add `disk_size_gib` to those two data-source blocks so the schema matches what's being set.

4. Bytes-only users can get clobbered on update. Because `memory_size_gib/disk_size_gib` are Optional + Computed and always populated on read (bytes / 1024^3, integer division), a user who sets only *_bytes with a non-GiB-aligned value gets a computed GiB in state; on a later update the GiB precedence wins and reverts the value to a GiB-rounded number. Please guard the precedence so it only applies when the GiB field was actually configured (and add a test for the bytes-update path).

5. Rebase onto master (2.4.2). The PR is based on 2.4.0 and conflicts. Note `checkForHotPlugChanges` was refactored — `memory_size_gib` change-tracking should go into `checkMemoryAndSocketsChange` alongside `memory_size_bytes`.

Testing to add/verify:

- Acceptance test for disk_size_gib create that asserts the resulting disk_size_bytes (the existing disk test only checks the attribute round-trips, not actual sizing).
- Read a disk-backed VM through data nutanix_virtual_machine_v2 / nutanix_virtual_machines_v2 and assert disk_size_gib.
- The bytes-only update path (item 4).
- Since the motivating issue (#961) is the 32-bit overflow, please confirm on both windows_386 and a 64-bit binary.
Happy to re-test once these are in. Thanks again!

PR #1088, opened 2026-02-23
add poweron Power off retries

### LINKED ISSUE #1089: Add retries for VM power on/off operations in Virtual Machine resource v2 to avoid ETag mismatch errors

When using the Virtual Machine resource v2 (`nutanix_virtual_machine_v2`), VM power on and power off operations during **Create** and **Update** can fail intermittently due to **ETag mismatch** errors. The Nutanix API uses the `If-Match` (ETag) header for optimistic concurrency; if the VM state changes between a read and the power operation, the request fails. This is especially common when the VM or cluster is busy, or when there is slight latency between create/update and the power state change.

For VM power on and VM power off operations in both **Create** and **Update** contexts:

- Add **4–5 retries** when the operation fails (e.g. due to ETag mismatch or transient errors).
- On each retry, **re-fetch the VM** to obtain a fresh ETag before calling the power API again, so that the `If-Match` header is always current and ETag mismatches are avoided when possible.

This makes create/update with power state changes more resilient without changing the resource schema or user-facing behavior.

**Describe alternatives you've considered**

- **Single attempt with no retry** – Leaves users exposed to intermittent ETag failures; they would need to re-run Terraform manually.
- **Retry without refreshing ETag** – Would not address ETag mismatch; the same stale ETag would be sent on retry and likely fail again.

**Additional context**

- **Resource:** `nutanix_virtual_machine_v2` (VMM v2).
- **Operations in scope:** Power On and Power Off during resource Create (when `power_state` is set) and during Update (when `power_state` is changed).
- **Goal:** Reduce flakiness and avoid ETag mismatch errors by retrying with a fresh ETag (e.g. 4–5 attempts with a short delay between attempts).

GullapalliAkhil:
### Summary

This PR adds **retry logic** for VM **power on** and **power off** operations in the Virtual Machine resource v2 (`nutanix_virtual_machine_v2`). When these operations fail (e.g. due to ETag mismatch or transient API issues), the provider now retries up to **5 times** with a **2.5 second** delay between attempts. On each retry, the VM is **re-read** to obtain a fresh ETag before calling the power API, which avoids ETag mismatch errors that commonly occur during Create and Update.

### Problem

- Power on/off during **Create** (when `power_state` is set) and **Update** (when `power_state` is changed) can fail with ETag mismatch if the VM state changes between the read and the power request.
- A single failed attempt forces users to re-run Terraform; retries with a fresh ETag make these operations more resilient.

### What's New

- **Create context:** When `power_state` is `ON` or `OFF`, power on/off is performed via `powerOnVMWithRetry` and `powerOffVMWithRetry` instead of a single API call. Each retry fetches the VM again for a fresh ETag.
- **Update context:** `callForPowerOnVM` and `callForPowerOffVM` now use the same retry helpers, so power state changes on update are also retried with fresh ETags.
- **Retry configuration:** Up to **5 attempts**, **2.5 s** delay between attempts; context cancellation is respected during waits.
- **Helper:** `extractTaskReferenceFromResponse` is introduced to obtain the task reference from the power API response in a type-safe way (used by both retry functions).

### Changes in This PR

- **`nutanix/services/vmmv2/resource_nutanix_virtual_machine_v2.go`**
  - **Create:** Replaced inline power on/off with `powerOnVMWithRetry` / `powerOffVMWithRetry`; removed the single pre-read used only for ETag (ETag is now fetched inside the retry loop).
  - **Update:** `callForPowerOnVM` and `callForPowerOffVM` now call `powerOnVMWithRetry` and `powerOffVMWithRetry` instead of a single power API call with a one-time ETag.
  - **New functions:**
    - `powerOnVMWithRetry(ctx, conn, vmID)` – retries power on with fresh ETag per attempt.
    - `powerOffVMWithRetry(ctx, conn, vmID)` – retries power off with fresh ETag per attempt.
    - `extractTaskReferenceFromResponse(resp)` – extracts `TaskReference` from the power API response (reflection-based for API response type).
  - Debug logging on retries and on success (e.g. task reference ExtId) for easier troubleshooting.

### Testing

- Existing acceptance and unit tests for the VM v2 resource should continue to pass; behavior is unchanged from the user’s perspective except for improved resilience on power operations.
- Manual testing: create/update VMs with `power_state` set to `ON` or `OFF` in environments where ETag mismatches have been observed; operations should succeed after retries when applicable.

### Backward compatibility

- No schema or attribute changes. Only the internal handling of power on/off is updated; existing configurations remain valid.

### DIFF
--- a/nutanix/services/vmmv2/resource_nutanix_virtual_machine_v2.go
+++ b/nutanix/services/vmmv2/resource_nutanix_virtual_machine_v2.go
@@ -1392,30 +1392,19 @@ func ResourceNutanixVirtualMachineV2Create(ctx context.Context, d *schema.Resour
 	}
 	d.SetId(utils.StringValue(uuid))
 
-	// read VM
-
-	readResp, errR := conn.VMAPIInstance.GetVmById(utils.StringPtr(*uuid))
-	if errR != nil {
-		return diag.Errorf("error while reading vm : %v", errR)
-	}
-	args := make(map[string]interface{})
-	args["If-Match"] = getEtagHeader(readResp, conn)
-
 	var PowerTaskRef import1.TaskReference
 	if powerState, ok := d.GetOk("power_state"); ok {
-		if powerState == "ON" {
-			resp, err := conn.VMAPIInstance.PowerOnVm(uuid, args)
-			if err != nil {
-				return diag.Errorf("error while powering on Virtual Machines : %v", err)
-			}
-			PowerTaskRef = resp.Data.GetValue().(import1.TaskReference)
-		} else if powerState == "OFF" {
-			resp, err := conn.VMAPIInstance.PowerOffVm(utils.StringPtr(d.Id()), args)
-			if err != nil {
-				return diag.Errorf("error while powering OFF : %v", err)
-			}
-			PowerTaskRef = resp.Data.GetValue().(import1.TaskReference)
-		}
+		switch powerState {
+    case "ON":
+      PowerTaskRef, err = powerOnVMWithRetry(ctx, conn, uuid)
+    case "OFF":
+      PowerTaskRef, err = powerOffVMWithRetry(ctx, conn, utils.StringPtr(d.Id()))
+    default:
+      return diag.Errorf("invalid power state: %s", powerState)
+    }
+    if err != nil {
+      return diag.Errorf("error while powering %s Virtual Machines: %v", powerState, err)
+    }
 	}
 	powertaskUUID := PowerTaskRef.ExtId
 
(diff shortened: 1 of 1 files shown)


GullapalliAkhil:
@Haroon-Dweikat-Ntx Tested the changes by inheriting the false cases?

Haroon-Dweikat-Ntx (approved):
LGTM

Haroon-Dweikat-Ntx:
> @Haroon-Dweikat-Ntx Tested the changes by inheriting the false cases?

Yes, I tested it by hardcode the wrong e-tag value as below: 

```
// Build args with fresh ETag for this attempt
args := make(map[string]interface{})
if attempt == maxRetries-1 {
	args["If-Match"] = getEtagHeader(readResp, conn)
} else {
	args["If-Match"] = "1234567890"
}
```

Haroon-Dweikat-Ntx:
[logs_OFF.log](https://github.com/user-attachments/files/25521731/logs_OFF.log)
[logs_ON.log](https://github.com/user-attachments/files/25521732/logs_ON.log)

nutanix/services/vmmv2/resource_nutanix_virtual_machine_v2.go
```diff
@@ -3175,6 +3164,143 @@ func expandPolicyReference(pr interface{}) *config.PolicyReference {
...
+	maxRetries := 5
+	retryDelay := 2500 * time.Millisecond // 2.5 seconds
+	var resp interface{}
+	var err error
+
+	for attempt := 0; attempt < maxRetries; attempt++ {
+		// Fetch VM to get latest ETag for each retry attempt
+		readResp, errR := conn.VMAPIInstance.GetVmById(vmID)
+		if errR != nil {
+			if attempt < maxRetries-1 {
```
GullapalliAkhil: Ignore the retries for GetVM call...

nutanix/services/vmmv2/resource_nutanix_virtual_machine_v2.go
```diff
@@ -3175,6 +3164,143 @@ func expandPolicyReference(pr interface{}) *config.PolicyReference {
...
+func powerOffVMWithRetry(ctx context.Context, conn *vmm.Client, vmID *string) (import1.TaskReference, error) {
+	maxRetries := 5
+	retryDelay := 2500 * time.Millisecond // 2.5 seconds
+	var resp interface{}
+	var err error
+
+	for attempt := 0; attempt < maxRetries; attempt++ {
+		// Fetch VM to get latest ETag for each retry attempt
+		readResp, errR := conn.VMAPIInstance.GetVmById(vmID)
+		if errR != nil {
```
GullapalliAkhil, line 3247: Same here ignore the retries for GetVM call

GullapalliAkhil (commented):
..

nutanix/services/vmmv2/resource_nutanix_virtual_machine_v2.go
Haroon-Dweikat-Ntx: why? each retry we need to read to call GetVM API, to get the latest e-tag value from the PC, actually that's why we adding the retry logic, because sometime power/off can fail with ETag mismatch if the VM state changes between the read and the power request.

GullapalliAkhil: Why to add retries for GetVM, there is no need of retrying if GetVM call failed. We need to retry only on power operations.

For every power ON, Power Off operation failed, GetVM Etag will be fetched every time. But if GetVM call is failed, no need of retrying it.

Haroon-Dweikat-Ntx (approved):
LGTM

PR #1104, opened 2026-03-18
fix(vmmv2): power state ETag handling and task-layer retry for nutanix_virtual_machine_v2

Haroon-Dweikat-Ntx:
Fixes power-on and power-off for `nutanix_virtual_machine_v2` when the API returns ETag mismatch errors (`If-Match` value does not match the current value on the server). Retries are consolidated at the **task layer** only: one API call per attempt, with up to `maxPowerRetries` attempts (10) and optional test retry behavior (wrong ETag on non-last attempts).

### Problem

- Power-on/power-off could fail with ETag mismatch because the wrong or stale `If-Match` header was sent.
- Retries were previously split between API-level and task-level, which complicated behavior and could still send the wrong ETag on every API retry.

### Solution

1. **Correct ETag usage**
   - Every power-on and power-off call uses the **exact** current ETag from the VM via `getEtagHeader(readResp, conn)` for the `If-Match` header (single GetVmById before each PowerOnVm/PowerOffVm).

2. **Task-layer retry only**
   - **Power-on:** `powerOnVM(ctx, conn, vmID, useCorrectEtag)` performs a single GetVmById + single PowerOnVm. `callForPowerOnVM` runs a loop (up to `maxPowerRetries`): read VM state → if not ON, call `powerOnVM` once → wait for task; on API or task failure, wait `powerTaskRetryDelay` (5s) and retry.
   - **Power-off:** Same pattern with `powerOffVM(ctx, conn, vmID, useCorrectEtag)` and `callForPowerOffVM`.

3. **Test retry (optional wrong ETag)**
   - `useCorrectEtag`: when `true`, send the correct ETag; when `false`, send a deliberately wrong ETag (e.g. last character incremented) so the task fails and the task-layer retry path is exercised.
   - In the task loop, only the **last** attempt uses the correct ETag: `useCorrectEtag := taskAttempt == maxPowerRetries-1`. Earlier attempts use the wrong ETag for testing.
   - Create path (initial power state) always calls with `useCorrectEtag = true`.

4. **Constants**
   - `maxPowerRetries = 10`
   - `powerTaskRetryDelay = 5 * time.Second`
   - Used for both power-on and power-off.

5. **Debug logging**
   - At start of each attempt: task attempt index, max attempts, and current VM `power_state`.
   - After task success: re-read VM and log actual `power_state`; log a WARNING if it does not match the desired state (ON/OFF).
   - Before returning final error: log "Power-on/Power-off failed after N attempts; last error: …".

### Files changed

- `nutanix/services/vmmv2/resource_nutanix_virtual_machine_v2.go`
  - `powerOnVM`: single call, `useCorrectEtag` parameter, wrong-ETag branch when false.
  - `powerOffVM`: single call, `useCorrectEtag` parameter, wrong-ETag branch when false.
  - `callForPowerOnVM`: task loop, one `powerOnVM` per attempt, pass `useCorrectEtag` based on attempt, retry on API or task failure.
  - `callForPowerOffVM`: task loop, one `powerOffVM` per attempt, same `useCorrectEtag` and retry behavior.
  - Create path: power-on/power-off use `powerOnVM(..., true)` / `powerOffVM(..., true)`.
  - Constants: `maxPowerRetries`, `powerTaskRetryDelay`.
  - Debug logs as above.

### Testing

- Run create/update with `power_state = "ON"` and `power_state = "OFF"` and confirm no ETag mismatch.
- With test retry enabled (wrong ETag on non-last attempts), confirm logs show retries and final success on the last attempt.

### DIFF
--- a/nutanix/services/vmmv2/resource_nutanix_virtual_machine_v2.go
+++ b/nutanix/services/vmmv2/resource_nutanix_virtual_machine_v2.go
@@ -3206,7 +3206,7 @@ func powerOnVM(ctx context.Context, conn *vmm.Client, vmID *string) (import1.Tas
 	}
 
 	args := make(map[string]interface{})
-   
+
 	args["If-Match"] = getEtagHeader(readResp, conn)
 
 	resp, err := conn.VMAPIInstance.PowerOnVm(vmID, args)
--- a/nutanix/services/vmmv2/resource_nutanix_virtual_machine_v2.go
+++ b/nutanix/services/vmmv2/resource_nutanix_virtual_machine_v2.go
@@ -1393,32 +1393,21 @@ func ResourceNutanixVirtualMachineV2Create(ctx context.Context, d *schema.Resour
 	}
 	d.SetId(utils.StringValue(uuid))
 
-	var PowerTaskRef import1.TaskReference
-	if powerState, ok := d.GetOk("power_state"); ok {
-		switch powerState {
-		case "ON":
-			PowerTaskRef, err = powerOnVMWithRetry(ctx, conn, utils.StringPtr(d.Id()))
-		case "OFF":
-			PowerTaskRef, err = powerOffVMWithRetry(ctx, conn, utils.StringPtr(d.Id()))
-		default:
-			return diag.Errorf("invalid power state: %s", powerState)
-		}
-		if err != nil {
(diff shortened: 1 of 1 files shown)


GullapalliAkhil (approved):

Haroon-Dweikat-Ntx:
this is duplciate of #1089

PR #1112, opened 2026-03-26
Mark user key secrets as sensitive and persist key_details on create

Haroon-Dweikat-Ntx:
### Summary

- Mark `api_key`, `secret_key`, and `access_key` fields in `nutanix_user_key_v2` as `Sensitive: true` so their values are redacted from Terraform plan/apply output and state CLI display.
- Set `key_details` in state during **Create** (where the API returns the full secret), instead of during **Read** (where the API no longer returns secrets). This ensures the one-time secret values are captured in state and available via `terraform output`.
- Add a `DEBUG`-level log of the created key response for troubleshooting.

### Motivation

The Nutanix IAM API only returns sensitive key material (`api_key`, `secret_key`, `access_key`) in the **create** response. Subsequent GET calls omit these fields. Previously, `key_details` was set only in the Read function, meaning the secrets were always empty in state after the initial apply.

Additionally, these fields were not marked as sensitive, so any value that did make it into state would appear in plaintext in plan/apply output.

### Changes

| File | Change |
|:-----|:-------|
| `nutanix/services/iamv2/resource_nutanix_user_key_v2.go` | Added `Sensitive: true` to `api_key`, `secret_key`, and `access_key` schema fields |
| `nutanix/services/iamv2/resource_nutanix_user_key_v2.go` | Moved `d.Set("key_details", ...)` from `Read` to `Create` so secrets are persisted from the create response |
| `nutanix/services/iamv2/resource_nutanix_user_key_v2.go` | Added debug log of created key response |

### DIFF
--- a/nutanix/services/iamv2/resource_nutanix_user_key_v2.go
+++ b/nutanix/services/iamv2/resource_nutanix_user_key_v2.go
@@ -125,8 +125,8 @@ func ResourceNutanixUserKeyV2() *schema.Resource {
 							Elem: &schema.Resource{
 								Schema: map[string]*schema.Schema{
 									"api_key": {
-										Type:     schema.TypeString,
-										Computed: true,
+										Type:      schema.TypeString,
+										Computed:  true,
 										Sensitive: true,
 									},
 								},
@@ -138,13 +138,13 @@ func ResourceNutanixUserKeyV2() *schema.Resource {
 							Elem: &schema.Resource{
 								Schema: map[string]*schema.Schema{
 									"secret_key": {
-										Type:     schema.TypeString,
-										Computed: true,
+										Type:      schema.TypeString,
+										Computed:  true,
 										Sensitive: true,
 									},
 									"access_key": {
-										Type:     schema.TypeString,
-										Computed: true,
+										Type:      schema.TypeString,
+										Computed:  true,
 										Sensitive: true,
 									},
 								},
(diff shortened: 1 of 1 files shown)


nutanix/services/iamv2/resource_nutanix_user_key_v2.go
```diff
@@ -290,9 +300,6 @@ func resourceNutanixUserKeyV2Read(ctx context.Context, d *schema.ResourceData, m
 	if err := d.Set("last_used_time", flattenTime(keyConfig.LastUsedTime)); err != nil {
 		return diag.Errorf("error while setting last_used_time: %v", err)
 	}
-	if err := d.Set("key_details", flattenKeyDetails(keyConfig.KeyDetails)); err != nil {
```
GullapalliAkhil, line 293: Incase of brownfield customers... what if API key is not set by create context and product team has fixed the bug to return the API key in response.

nutanix/services/iamv2/resource_nutanix_user_key_v2.go
```diff
@@ -125,8 +125,9 @@ func ResourceNutanixUserKeyV2() *schema.Resource {
 							Elem: &schema.Resource{
 								Schema: map[string]*schema.Schema{
 									"api_key": {
-										Type:     schema.TypeString,
-										Computed: true,
+										Type:      schema.TypeString,
+										Computed:  true,
```
GullapalliAkhil, line 129: If it is sensitive, can it be read by customer?

Haroon-Dweikat-Ntx: Yes, sensitive mean its will not be logged out in plan or output as planned texted 
but it can be used and exists in the state file

nutanix/services/iamv2/resource_nutanix_user_key_v2.go
Haroon-Dweikat-Ntx: This mean that the api behavior changed, so we need to modify the resource with new changes

GullapalliAkhil (approved):

PR #1168, opened 2026-05-14
Release/2.4.3

### LINKED ISSUE #1176: NGT ISO INSERT eject should be a no-op if iso eject is already done as part of NGT installation, custom eject.
User inserted NGT ISO, NGT got installed (which auto-ejects the ISO from the CD-ROM), then upgraded to the latest provider version, then runs terraform destroy.
cdrom_ext_id becomes empty after Read (no GuestTools CD-ROM found on the VM). This causing the Terraform destroy to fail.

Expected: Delete returns a warning and skips the eject gracefully.

GullapalliAkhil:
Fix:
   NGT ISO INSERT eject should be a no-op if iso eject is already done as part of NGT installation, custom eject.
   
 
 Tests:
Upgrade scenario — Insert + NGT Installed → Destroy

User inserted NGT ISO, NGT got installed (which auto-ejects the ISO from the CD-ROM), then upgraded to the latest provider version, then runs terraform destroy.
cdrom_ext_id becomes empty after Read (no GuestTools CD-ROM found on the VM).
Expected: Delete returns a warning and skips the eject gracefully.
Upgrade scenario — Insert + No Installation → Destroy

User inserted NGT ISO but never installed NGT (ISO is still mounted in the CD-ROM), then upgraded to the latest provider version, then runs terraform destroy.
cdrom_ext_id gets populated from the VM's CD-ROM with iso_type == GUEST_TOOLS.
Expected: Delete calls ejectCdromISO successfully and ejects the ISO.
Greenfield — Insert + NGT Installed → Destroy

Fresh setup with latest provider. User inserts ISO, NGT installs (auto-ejects ISO), then runs terraform destroy.
cdrom_ext_id is in state from Create (stale, but still valid CD-ROM device).
Expected: EjectCdRomById API handles this as a no-op since the CD-ROM is already empty.
Greenfield — Insert + No Installation → Destroy

Fresh setup with latest provider. User inserts ISO, does NOT install NGT, then runs terraform destroy.
cdrom_ext_id is valid and ISO is still mounted.
Expected: Delete calls ejectCdromISO and the API ejects the ISO successfully.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+## 2.4.3 (May 15, 2026)
+[Full Changelog](https://github.com/nutanix/terraform-provider-nutanix/compare/v2.4.2...v2.4.3)
+
+**Fixed Bugs:**
+- `nutanix_ngt_insert_iso_v2`: Destroy/eject fails when `cdrom_ext_id` is null after provider upgrade (ISO was already ejected during NGT installation).
+- Typo fix in doc strings and README. [#1140](https://github.com/nutanix/terraform-provider-nutanix/pull/1140)
+
+
 ## 2.4.2(April 15, 2026) 
 [Full Changelog](https://github.com/nutanix/terraform-provider-nutanix/compare/v2.4.0...v2.4.2)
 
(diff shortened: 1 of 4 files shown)


GullapalliAkhil:
2.4.3 is rolled out as beta release to unblock customer.

Issue #1139, opened 2026-04-16
VM resource attempts to remove externally attached Volume Group disks on every plan/apply


rubensix1:
### Nutanix Cluster Information
Version pc.7.5.0.1
NCC Version: ncc-5.3.0

### Terraform Version
registry.terraform.io/nutanix/nutanix v2.4.2

### Affected Resource(s)

nutanix_virtual_machine_v2
nutanix_volume_group_v2
nutanix_volume_group_disk_v2
nutanix_volume_group_vm_v2

### Terraform Configuration Files
```
resource "nutanix_virtual_machine_v2" "vm"{
    name= "example-vm-1"
    num_cores_per_socket = 1
    power_state = "OFF"
    num_sockets = 1
    cluster {
        ext_id = "CLUSTER_ID"
    }

    disks{
      disk_address{
          bus_type = "SCSI"
          index = 0
      }
      backing_info{
        vm_disk{
           disk_size_bytes = "1073741824"
           storage_container{
              ext_id = "STORAGE_CONTAINER_ID"
           }
        }
      }
  }


}

resource "nutanix_volume_group_v2" "volume_group" {
  name              = "reproduce-disks"
  cluster_reference = "CLUSTER_ID"
}

resource "nutanix_volume_group_disk_v2" "disks" {
  depends_on          = [nutanix_volume_group_v2.volume_group]
  volume_group_ext_id = nutanix_volume_group_v2.volume_group.ext_id
  disk_size_bytes     = "1073741824"
  disk_data_source_reference {
    ext_id      = "STORAGE_CONTAINER_ID"
    entity_type = "STORAGE_CONTAINER"
  }
}

resource "nutanix_volume_group_vm_v2" "attach_disk" {
  depends_on = [
    nutanix_volume_group_v2.volume_group,
    nutanix_volume_group_disk_v2.disks,
    nutanix_virtual_machine_v2.vm
  ]
  volume_group_ext_id = nutanix_volume_group_v2.volume_group.id
  vm_ext_id           = nutanix_virtual_machine_v2.vm.id
  index               = 1
}
```

A plan / apply (with Volume Group Disk attached) after the VM is created
```
  # nutanix_virtual_machine_v2.vm will be updated in-place
  ~ resource "nutanix_virtual_machine_v2" "vm" {
        id                           = "be533fba-9e9a-43fc-484e-d89f80119639"
        name                         = "example-vm-1"
        # (24 unchanged attributes hidden)

[...]

PR #1212, opened 2026-07-24
Fix 1079 & 1139

### LINKED ISSUE #1079: Attaching volume group to VM invalidates state
PC/PE: 7.3
### Terraform Version
1.12
### Affected Resource(s)
* nutanix_volume_group_vm_v2
* nutanix_virtual_machine
* nutanix_virtual_machine_v2
```hcl
resource "nutanix_volume_group_v2" "volume_group" {
  name                               = "vg-${random_pet.vg-name.id}"
  should_load_balance_vm_attachments = false
  sharing_status                     = "SHARED"

  cluster_reference = local.cluster_extId

  is_hidden = false

}

resource "nutanix_volume_group_disk_v2" "volume_group_disk" {
  volume_group_ext_id = nutanix_volume_group_v2.volume_group.ext_id
  index               = 1
  disk_size_bytes     = provider::units::from_gb(2)

  disk_data_source_reference {
    ext_id      = data.nutanix_storage_containers_v2.storage_container_ext_id.storage_containers[0].ext_id
    entity_type = "STORAGE_CONTAINER"
  }

  lifecycle {
    ignore_changes = [
      disk_data_source_reference
    ]
  }

}

resource nutanix_virtual_machine {
# Any VM configuration
}

resource "nutanix_volume_group_vm_v2" "vg_vm_attachment" {
  volume_group_ext_id = nutanix_volume_group_v2.volume_group.ext_id
  vm_ext_id           = module.vm.vm.id
}
```
### Expected Behavior
A virtual machine and volume group are deployed and the volume group is attached to the virtual machine
### Actual Behavior
A virtual machine and volume group are depoyed and attached, but on subsequent Terraform plan and apply actions, Terraform attempts to remove the volume group from the VM because the resulting disk was not part of the initial VM configuration.
### Steps to Reproduce
1. `terraform apply`
### Important Factors
This can be worked around by specifying the volume group directly on a ```nutanix_virtual_machine_v2``` and avoiding ```nutanix_volume_group_vm_v2```, however there does not appear to be a workaround for ```nutanix_virtual_machine```.

kvandenhoute:
Detach volumegroup that is attached on vm, before destroying the vm. Prismc does not allow to delete vm with attached volumegroup

fixes #1079 & #1139

### DIFF
--- a/nutanix/services/vmmv2/resource_nutanix_virtual_machine_v2.go
+++ b/nutanix/services/vmmv2/resource_nutanix_virtual_machine_v2.go
@@ -2187,10 +2187,22 @@ func ResourceNutanixVirtualMachineV2Update(ctx context.Context, d *schema.Resour
 
 func ResourceNutanixVirtualMachineV2Delete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	conn := meta.(*conns.Client).VmmAPI
+	taskconn := meta.(*conns.Client).PrismAPI
 
 	readResp, err := conn.VMAPIInstance.GetVmById(utils.StringPtr(d.Id()))
 	if err != nil {
 		return diag.Errorf("error while reading vm : %v", err)
+	}
+
+	if detachErr := detachADSFVolumeGroupReferences(ctx, d, meta, conn, readResp.Data.GetValue().(config.Vm)); detachErr != nil {
+		return detachErr
+	}
+
+	// Detaching ADFS-backed disks mutates the VM and changes its ETag.
+	// Re-read the VM so DeleteVmById uses the latest If-Match value.
+	readResp, err = conn.VMAPIInstance.GetVmById(utils.StringPtr(d.Id()))
+	if err != nil {
+		return diag.Errorf("error while refreshing vm before delete : %v", err)
 	}
 	// Extract E-Tag Header
 	args := make(map[string]interface{})
@@ -2204,8 +2216,6 @@ func ResourceNutanixVirtualMachineV2Delete(ctx context.Context, d *schema.Resour
 	taskUUID := TaskRef.ExtId
 
 	// calling group API to poll for completion of task
-
-	taskconn := meta.(*conns.Client).PrismAPI
 	// Wait for the task to complete
 	stateConf := &resource.StateChangeConf{
 		Pending: []string{"PENDING", "RUNNING", "QUEUED"},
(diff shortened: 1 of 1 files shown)


kvandenhoute:
it seems like linting is not ok on main branch

kvandenhoute:
@GullapalliAkhil @Haroon-Dweikat-Ntx Please let me know if any changes are required

Haroon-Dweikat-Ntx:
Hi @kvandenhoute, thank you for taking the time to work on this and submit a PR to address these issues!

While we understand the underlying problem with the VM deletion failing due to attached Volume Groups, we won't be able to merge this specific approach.

Adding the automatic disassociation logic to the VM deletion path goes against standard Terraform guidelines. Relationships and attachments should be managed by their specific attachment resources to maintain proper state and lifecycle management, rather than building that logic into the core VM resource itself.

To manage Volume Group attachments correctly, users should use the dedicated `nutanix_volume_group_vm_v2` resource instead of configuring it directly within the VM resource. Using `nutanix_volume_group_vm_v2` ensures that Terraform can properly handle the attach and detach operations independently in the dependency graph during apply and destroy operations.

Because of this, I'll be closing this PR, but we really appreciate your contribution and effort here!

kvandenhoute:
Hi @Haroon-Dweikat-Ntx,

I feel you are contradicting yourself? You suggest us to use the adfs_volume_group_reference directly on the vm resource here: https://github.com/nutanix/terraform-provider-nutanix/issues/1139#issuecomment-4969765038

In this comment you literally say to not split the attach from the vm: 

> "When the VM is managed by Terraform, attach the volume group inside the VM's disks block using backing_info.adfs_volume_group_reference, and do not use nutanix_volume_group_vm_v2 for that attachment. This keeps a single source of truth and eliminates the drift."

This PR fixes that you can actually do that, because this feature is currently bugged, confirmed on the other PR too: https://github.com/nutanix/terraform-provider-nutanix/issues/1079#issuecomment-5000140714

Our original solution is to use the "nutanix_volume_group_vm_v2", which is exactly why we opened the PR: it does not work. 
Because you then commented that you don't want us to use it and instead do it directly on the vm, I put work into fixing this. 

Now you say you can not actually use this feature (then why is it available even?), and we should go back to our original solution, which does not work.

Please, with terraform provider 2.4.2 try yourself to create:
a volumegroup with atleast 1 disk
a vm that uses that volumegroup (so attach it) and cloud init

Now try:
a. changing something on the cloud-init triggering a recreate of the VM
b. destroying the vm
c. changing something non-replace on the vm
d. changing nothing on the vm

Neither of the two solutions work for all A,B,C,D. 

Using the adfs_volume_group_reference on nutanix_vm_v2 won't allow A/B, it does work for C/D
Using the nutanix_volume_group_vm_v2 allows A/B but C/D will remove the attachment

Please reinvestigate these issues, as currently the provider is not workable with volumegroups. 


I don't agree that this is anti-terraform, but if you'd like a detach flag could be added;
 nutanix_vm_v2 : force_detach_disks_on_recreate <boolean> or w/e
 
 
Please do not close this.

Haroon-Dweikat-Ntx:
Hi @kvandenhoute,

First, I want to sincerely apologize for the mixed messaging here. You are absolutely right to call out the contradiction, and I completely understand how frustrating it is to be sent back and forth, especially after you put in the time and effort to build a fix for us.

You have summarized the current broken state perfectly:
- **Inline (`adfs_volume_group_reference`):** Fails on A/B (Destroy/Recreate) because the API rejects VM deletion with attached VGs.
- **Standalone (`nutanix_volume_group_vm_v2`):** Fails on C/D (In-place updates/No-op) because the VM resource incorrectly sees the external attachment as drift and removes it.

When looking at this from a Terraform architecture perspective, the standalone resource (`nutanix_volume_group_vm_v2`) is the correct design pattern. My earlier advice on #1139 (to use inline) was a workaround attempt that ultimately led you down a path blocked by the API deletion constraints, which this PR valiantly tried to patch.

The good news is that we are actively working on a fix for the underlying drift issue, targeted for our next release (2.5.0). With this fix, the VM resource will no longer show a diff on `apply` or `plan` when a Volume Group is attached via the standalone resource, allowing scenarios A, B, C, and D to all work correctly.

Thank you again for holding us accountable and for your deep dive into these use cases. We really appreciate your patience while we get this properly resolved.

kvandenhoute:
@Haroon-Dweikat-Ntx Ok that is great to hear, I'm looking forward to it being resolved.

I do fail to understand why the inline adfs_volume_group_reference then exists, if its purpose is not to use it. Maybe it should be removed, or documented that this feature is not advised to be used.

PR #1220, opened 2026-08-13
[Bug] Enforce TLS certificate verification in v4 SDK clients; fix provider panics

### LINKED ISSUE #1099: Segfault on flattenDisk, v1.9.5, v2.4.0

Plugin crashes, not relevant.

### Terraform Version

```
Terraform v1.14.6
on darwin_arm64
```

### Affected Resource(s)

Plugin itself crashes, not relevant.

Plugin crashes, not relevant.

### Panic Output

https://gist.github.com/Vaelatern/83b5fa726b2790362649121d142323b4

PrazwalR:
## Summary

Fixes a TLS certificate-verification bypass affecting every `*_v2` resource and
data source, plus a family of provider crashes and swallowed-error bugs found
alongside it.

⚠️ **Contains a breaking change.** See [Breaking Changes](#breaking-changes).

88 files changed. Production code is +821/−558 across 80 files; the remainder is
945 lines of new tests and the removal of `out.txt` (a 1,373-line debug log from
2019 committed at the repo root).

## Problem

`ConfigureV4Client` hardcoded `VerifySSL: false`:

```go
cfg := &V4ApiClientConfig{
    Host:      credentials.Endpoint,
    Port:      port,
    VerifySSL: false,          // credentials.Insecure never read
    ...
}
```

All 13 v4 SDK clients consumed this, so certificate verification was disabled
for the entire v4 surface regardless of the provider's `insecure` setting —
which defaults to `false` and is documented as meaning "verify". There was no
way for an operator to turn verification on.

A party on the network path could present any certificate and capture the
`Authorization: Basic` header or `X-Ntnx-Api-Key`, and could alter responses so
`terraform plan` and the cluster's actual state diverge.

`git log -S` shows this dates from the introduction of the v4 clients; a later
refactor centralized the constant without revisiting it.

Investigating the client layer surfaced the further issues listed below.

## Solution

**Security**
- `VerifySSL: !credentials.Insecure`.
- Each client builds its own `http.Client` and `Transport` (cloned from
  `http.DefaultTransport`). Previously all clients shared `http.DefaultClient`,
  so one client's TLS/proxy settings applied process-wide.
- Removed log statements emitting cluster passwords
  (`nutanix_password_change_request_v2`) and the iSCSI CHAP secret
  (`nutanix_volume_group_v2`).

**Crashes**
- `flattenOneOfOvaSource` asserted `UrlSource`/`VmDiskSource`; `OneOfOvaSource`
  only holds `OvaUrlSource`/`OvaVmSource`/`ObjectsLiteSource`. It also read
  `ExtId` rather than `VmExtId` and emitted a key absent from the schema —
  the branch had never executed. Now uses comma-ok and the correct fields.
- `flattenDisk` dereferenced `DeviceProperties.DiskAddress` and
  `StorageConfig.StorageContainerReference` without nil checks (#1099). Both are
  optional for CD-ROM and volume-group-backed disks.
- ~73 enum lookups did `m[k]` then `.(int)`, panicking on a miss. An omitted
  optional attribute in a nested block arrives as `""`, as does `d.Get` on an
  unset optional, so this was reachable from valid configuration.
- `filter()` and `CheckResponse` used unchecked type assertions on remote data
  and discarded a JSON unmarshal error.
- Replaced `log.Fatal`/`log.Fatalf` (4 sites), which call `os.Exit` and kill the
  plugin process mid-apply with no diagnostic and no state write.

**Error handling**
- A non-2xx response with an empty body is now an error. It returned `nil`, so
  failed `DELETE` calls (12 call sites pass a nil response target) were reported
  as success and Terraform dropped still-existing resources from state.
- Added `client.ErrNotFound` for 404s. `nutanix_karbon_private_registry` Read and
  Exists treated *any* error as deletion, so a transient 500 removed the resource
  from state.
- Fixed ~24 sites reporting or checking the wrong error variable — losing the
  real API message (`%!v(<nil>)`) or ignoring `d.Set` failures entirely.
- Session-auth failures now surface instead of returning a cookie-less client.
- Fixed 8 flatteners using `make([]T, len(x))` + `append`, which prepended N
  empty entries to lists in state.
- Disks with no UUID are no longer silently dropped from VM state.

**Tooling**
- `.golangci.yml` migrated to the v2 schema. The previous file failed schema
  validation, so `make lint` could not run at all. Auto-migration was not
  possible (the v1 content referenced linters removed years ago), so it was
  hand-ported preserving the original intent. Linters added to golangci-lint
  since the file was last updated are disabled in a commented block for
  follow-up; `makezero` is left **enabled** since this PR fixes that class.
- Resolved merge conflict markers committed in `.env.example`.
- Removed `out.txt`.
- Gitignored `object_store_cert.json`, which the test suite writes to the repo
  root.

## Security

| Finding | Severity | Fix |
|---|---|---|
| TLS verification disabled on all v4 clients | Critical | Honour `insecure` |
| Shared global HTTP client leaks TLS/proxy config | High | Per-client transport |
| Cluster passwords written to debug log | High | Log removed |
| iSCSI CHAP secret written to debug log | High | Log removed |
| Read failure treated as resource deletion | High | `ErrNotFound` sentinel |

## Testing

```
go build ./...          PASS
go vet ./...            PASS
scripts/gofmtcheck.sh   PASS
make test               PASS  (32 packages)
go test ./utils/...     PASS
go test -race           PASS  (touched packages)
golangci-lint run       PASS  (0 issues)
go install .            PASS
```

35 new tests across 6 files. Each of the three highest-severity fixes was
individually reverted to confirm the corresponding test fails:

- `VerifySSL` reverted → `FAIL: VerifySSL = false, want true (insecure=false)`
- `flattenDisk` guard reverted → `panic: invalid memory address or nil pointer dereference`
- OVA assertion reverted → `FAIL: TestFlattenOneOfOvaSource_UrlSource`

`TestCheckResponse_404PreservesEntityNotFoundText` pins a compatibility contract
worth calling out: around thirty v3 resources detect deletion with
`strings.Contains(err, "ENTITY_NOT_FOUND")`, and their fallback branch calls
`Delete` on the resource. Adding `ErrNotFound` therefore has to keep the raw API
body inside the error text, and that test proves both mechanisms hold at once.

**Not run:** the acceptance suite, which requires live Prism Central, Foundation
and NDB instances. **The TLS change should get one acceptance run against real
hardware before release**, since it alters connection behavior for every `*_v2`
resource. I have proven the config value is computed correctly, not that the SDK
honours it end to end.

## Breaking Changes

**TLS certificate verification is now enforced for v4 (`*_v2`) resources.**

Deployments whose Prism Central presents a self-signed or otherwise untrusted
certificate will see:

```
x509: certificate signed by unknown authority
```

Resolve by installing a trusted certificate, or opt out explicitly:

```hcl
provider "nutanix" {
  insecure = true   # or NUTANIX_INSECURE=true
}
```

This restores the previous connection behavior, now as a deliberate choice.

Documented in `CHANGELOG.md` and `website/docs/index.html.markdown`.

## Other behavior changes

Neither is a defect, but both are judgement calls worth a reviewer's attention:

1. **Empty-body error responses now fail.** This surfaces previously-silent
   failures, most importantly failed `DELETE` operations that were reported as
   success. Users may start seeing errors on operations they believed were
   working — those operations were already failing.
2. **`iscsi_features.target_secret = ""` is now omitted** rather than sent as an
   empty string. More correct, but if the API distinguishes "omitted" from
   "empty", it removes a way to clear a CHAP secret. The field is `Computed` and
   documented as unretrievable once set, so clearing is not a supported
   operation.

## Notes

Deliberately **not** included, to keep this reviewable:

- ~500 further `GetValue().(T)` assertions across the v2 services. Same class,
  but each needs a judgement call on the fallback. Find them with:
  `grep -rn "GetValue()\.(" --include="*.go" nutanix/services/`
- Re-enabling `errcheck` — hundreds of unchecked `d.Set` returns, each able to
  produce silently wrong state.
- Re-enabling `staticcheck` / `gosec`.
- The v3 client passes `context.TODO()` throughout, so operations are not
  cancellable.
- `resource_nutanix_password_manager_v2.go` rebuilds a Prism client after a
  password change but drops `APIKey` and `CustomHeaders`, so the retry path fails
  behind Cloudflare Access. Pre-existing; needs that environment to test.

I am happy to split this into separate PRs (security / crashes / tooling) if that
would be easier to review.

Fixes #1099.


nutanix/services/clustersv2/resource_nutanix_cluster_discover_unconfigured_nodes_v2.go
```diff
@@ -212,9 +212,10 @@ func DatasourceNutanixClusterDiscoverUnconfiguredNodesV2Create(ctx context.Conte
...
 			}
-			pVal := subMap[addressType.(string)]
-			p := config.AddressType(pVal.(int))
-			body.AddressType = &p
+			if pVal, ok := subMap[addressType.(string)].(int); ok {
+				p := config.AddressType(pVal)
+				body.AddressType = &p
+			}
 		}
 	}
```
Haroon-Dweikat-Ntx:
```suggestion
    if addressType, ok := d.GetOk("address_type"); ok {
		body.AddressType = common.ExpandEnum[config.AddressType](addressType.(string))
	}
```

nutanix/services/dataprotectionv2/resource_nutanix_recovery_points_v2.go
```diff
@@ -323,19 +323,21 @@ func ResourceNutanixRecoveryPointsV2Create(ctx context.Context, d *schema.Resour
...
 			"COMPLETE": two,
 		}
-		pVal := statusMap[status.(string)]
-		p := common.RecoveryPointStatus(pVal.(int))
-		body.Status = &p
+		if pVal, ok := statusMap[status.(string)].(int); ok {
+			p := common.RecoveryPointStatus(pVal)
+			body.Status = &p
+		}
 	}
```
Haroon-Dweikat-Ntx:
```suggestion
	if status, ok := d.GetOk("status"); ok {
		body.Status = commonUtils.ExpandEnum[common.RecoveryPointStatus](status.(string))
	}
```

nutanix/services/dataprotectionv2/resource_nutanix_recovery_points_v2.go
```diff
@@ -701,9 +703,10 @@ func expandApplicationConsistentProperties(appConsistentProp interface{}) (*conf
...
 			"COPY_BACKUP": three,
 		}
-		pVal := backupTypeMap[backupType.(string)]
-		p := common.BackupType(pVal.(int))
-		appConsistentPropObj.BackupType = &p
+		if pVal, ok := backupTypeMap[backupType.(string)].(int); ok {
+			p := common.BackupType(pVal)
+			appConsistentPropObj.BackupType = &p
+		}
 	}
```
Haroon-Dweikat-Ntx:
```suggestion
	if backupType, ok := appConsistentPropVal["backup_type"]; ok {
		appConsistentPropObj.BackupType = commonUtils.ExpandEnum[common.BackupType](backupType.(string))
	}
```

PR #1222, opened 2026-08-16
resource/nutanix_virtual_machine_v2: wait for a routable IP after power-on (#871)

### LINKED ISSUE #871: nutanix_virtual_machine_v2 returns APPIPA
AOS 7
Prism pc.2024.3.1.1

### Terraform Version

Terraform 1.12.1

### Affected Resource(s)

* nutanix_virtualmachine_v2

```hcl

nutanix_virtual_machine_v2.vm[count.index].nics[0].network_info[0].ipv4_info[0].learned_ip_addresses[0].value

# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file.
```
### Expected Behavior
This should return the ip address

### Actual Behavior
Returns APPIPA sometimes as it registers the address to soon

### Workarround
```hcl
length([
    for ip in nutanix_virtual_machine_v2.vm[count.index].nics[0].network_info[0].ipv4_info[0].learned_ip_addresses :
    ip.value if !can(regex("^169\\.254\\.", ip.value))
  ]) > 0 ?
    tolist([
      for ip in nutanix_virtual_machine_v2.vm[count.index].nics[0].network_info[0].ipv4_info[0].learned_ip_addresses :
      ip.value if !can(regex("^169\\.254\\.", ip.value))
    ])[0] :
    null
```

1. `terraform apply`
### Important Factors
DHCP is running on a Windows DHCP server.

* #0000

andrew-sumner:
Closes #871.

## Problem

In real deployments, `nutanix_virtual_machine_v2` **intermittently returns the wrong IP on create** —
an APIPA / link-local address (`169.254.0.0/16`) instead of the VM's actual DHCP address. Nutanix
guest tools surface the link-local address **transiently before DHCP completes**, and
`getFirstIPAddress` returned the first learned address unconditionally. As a result:

- `d.SetConnInfo` (and the IP used by provisioners) could be a `169.254.x` address, and
- the IP-wait state machine treated the APIPA as "available" and stopped waiting.

Consumers that need a routable address (provisioners, modules) then fail intermittently — it is a
race against DHCP latency, so it shows up sporadically on slow-lease networks.

## Fix

Make the post-power-on IP wait skip APIPA/link-local and keep polling for a routable address. Two
new arguments, modeled on the vSphere provider's guest-network waiter:

| Argument | Type | Default | Behavior |
|---|---|---|---|
| `wait_for_ip_timeout` | int (minutes) | `5` | How long to wait after power-on for a usable IPv4 address; `<1` disables. Mirrors vSphere `wait_for_guest_net_timeout`. |
| `wait_for_ip_routable` | bool | `true` | Ignore `169.254.0.0/16` while waiting. Set `false` to accept the first learned address (previous behavior). Mirrors vSphere `wait_for_guest_net_routable`. |

`isAPIPA()` (IPv4 link-local only) is used by both `getFirstIPAddress` and `waitForIPRefreshFunc`.

## Compatibility

- Defaults match the vSphere provider exactly (`5` min, routable `true`, `<1` disables).
- The IP wait timing out is non-fatal (logged; the resource still completes), so an intentionally
  APIPA-only VM simply waits out the timeout instead of failing.
- `wait_for_ip_routable = false` restores the prior "accept first learned address" behavior.

## Tests

**Unit** (`resource_nutanix_virtual_machine_v2_apipa_test.go`):

- `TestIsAPIPA` — `169.254.x` true; routable / loopback / IPv6 link-local / empty / malformed false.
- `TestGetFirstIPAddress` — apipa-then-routable → routable; apipa-only (routable) → skipped;
  apipa-only (not routable) → accepted; static fallback; no-network-info → empty.

**End-to-end** on an AHV cluster. The DHCP race was made deterministic from guest cloud-init (a
netplan bringing the NIC up on a link-local address with `dhcp4: false`, then a timer switching to
DHCP ~90 s later), reproducing the real timeline:

| Time | NIC learned IPs |
|---|---|
| t+24 s | `169.254.x` (link-local only) |
| t+136 s | routable address (link-local gone) |

Same VM configuration, only the provider version changed:

| | current release | this PR |
|---|---|---|
| waits for a routable IP | no | yes (≤ `wait_for_ip_timeout`, ignoring `169.254.x`) |
| IP reported | `169.254.x` | routable |
| consumer gating on a routable IP | fails | succeeds |

## Docs

`website/docs/r/virtual_machine_v2.html.markdown` documents both new arguments.

### DIFF
--- a/nutanix/services/vmmv2/resource_nutanix_virtual_machine_v2.go
+++ b/nutanix/services/vmmv2/resource_nutanix_virtual_machine_v2.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
+	"net"
 	"reflect"
 	"strconv"
 	"time"
@@ -614,6 +615,18 @@ func ResourceNutanixVirtualMachineV2() *schema.Resource {
 				Optional:     true,
 				Default:      "ON",
 				ValidateFunc: validation.StringInSlice([]string{"ON", "OFF", "PAUSED", "UNDETERMINED"}, false),
+			},
+			"wait_for_ip_timeout": {
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Default:     5,
+				Description: "Minutes to wait after power-on for the VM to report a usable IPv4 address. A value less than 1 disables the wait. Modeled on the vSphere provider's wait_for_guest_net_timeout.",
+			},
+			"wait_for_ip_routable": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     true,
+				Description: "When waiting for an IP (see wait_for_ip_timeout), require a routable address, ignoring APIPA/link-local (169.254.0.0/16) addresses reported transiently before DHCP completes. Set false to accept the first learned address, including APIPA.",
 			},
 			"vtpm_config": {
 				Type:     schema.TypeList,
@@ -1409,13 +1422,14 @@ func ResourceNutanixVirtualMachineV2Create(ctx context.Context, d *schema.Resour
 	nics := d.Get("nics")
 	hasNics := nics != nil && len(common.InterfaceToSlice(nics)) > 0
 
-	if d.Get("power_state") == "ON" && hasNics {
+	waitForIPTimeout := d.Get("wait_for_ip_timeout").(int)
+	if d.Get("power_state") == "ON" && hasNics && waitForIPTimeout >= 1 {
 		// Wait for the VM to be available
 		waitIPConf := &resource.StateChangeConf{
 			Pending:    []string{"WAITING"},
 			Target:     []string{"AVAILABLE"},
-			Refresh:    waitForIPRefreshFunc(ctx, conn, utils.StringValue(uuid)),
-			Timeout:    timeout,
+			Refresh:    waitForIPRefreshFunc(ctx, conn, utils.StringValue(uuid), d.Get("wait_for_ip_routable").(bool)),
+			Timeout:    time.Duration(waitForIPTimeout) * time.Minute,
 			Delay:      delay,
 			MinTimeout: delay,
 		}
@@ -1427,7 +1441,7 @@ func ResourceNutanixVirtualMachineV2Create(ctx context.Context, d *schema.Resour
 			vmResp := vm.Data.GetValue().(config.Vm)
 
 			if len(vmResp.Nics) > 0 && vmResp.Nics[0].NetworkInfo != nil {
-				ipAddr := getFirstIPAddress(vmResp.Nics[0])
+				ipAddr := getFirstIPAddress(vmResp.Nics[0], d.Get("wait_for_ip_routable").(bool))
 				if ipAddr != "" {
 					d.SetConnInfo(map[string]string{
 						"type": "ssh",
@@ -3346,16 +3360,26 @@ func resourceNutanixVirtualMachineV2StateUpgradeV0(ctx context.Context, rawState
 	return rawState, nil
 }
 
+// isAPIPA reports whether s is an IPv4 link-local / APIPA address (169.254.0.0/16).
+// Nutanix guest tools report these transiently before DHCP assigns a real lease, so a
+// caller waiting for a usable address must skip them and keep polling (issue #871).
+func isAPIPA(s string) bool {
+	ip := net.ParseIP(s)
+	return ip != nil && ip.To4() != nil && ip.IsLinkLocalUnicast()
+}
+
 // getFirstIPAddress returns the first available IP address from a NIC.
 // It checks both DHCP learned IPs and statically configured IPs.
-func getFirstIPAddress(nic config.Nic) string {
+func getFirstIPAddress(nic config.Nic, routable bool) string {
 	if nic.NetworkInfo == nil {
 		return ""
 	}
 	// Check for DHCP learned IPs first
-	if nic.NetworkInfo.Ipv4Info != nil && len(nic.NetworkInfo.Ipv4Info.LearnedIpAddresses) > 0 {
-		if nic.NetworkInfo.Ipv4Info.LearnedIpAddresses[0].Value != nil {
-			return *nic.NetworkInfo.Ipv4Info.LearnedIpAddresses[0].Value
+	if nic.NetworkInfo.Ipv4Info != nil {
+		for _, ip := range nic.NetworkInfo.Ipv4Info.LearnedIpAddresses {
+			if ip.Value != nil && (!routable || !isAPIPA(*ip.Value)) {
+				return *ip.Value
+			}
 		}
 	}
 	// Check for statically configured IP
@@ -3365,7 +3389,7 @@ func getFirstIPAddress(nic config.Nic) string {
 	return ""
 }
 
-func waitForIPRefreshFunc(ctx context.Context, client *vmm.Client, vmUUID string) resource.StateRefreshFunc {
+func waitForIPRefreshFunc(ctx context.Context, client *vmm.Client, vmUUID string, routable bool) resource.StateRefreshFunc {
 	return func() (interface{}, string, error) {
 		getVmByIdRequest := import3.GetVmByIdRequest{
 			ExtId: utils.StringPtr(vmUUID),
@@ -3383,8 +3407,11 @@ func waitForIPRefreshFunc(ctx context.Context, client *vmm.Client, vmUUID string
 					// Check for DHCP learned IPs
 					if nic.NetworkInfo.Ipv4Info != nil {
 						for _, ip := range nic.NetworkInfo.Ipv4Info.LearnedIpAddresses {
-							if ip.Value != nil {
+							if ip.Value != nil && (!routable || !isAPIPA(*ip.Value)) {
 								return resp, "AVAILABLE", nil
+							}
+							if ip.Value != nil && routable && isAPIPA(*ip.Value) {
+								log.Printf("[DEBUG] VM %s: ignoring APIPA/link-local %q; waiting for a routable IP", vmUUID, *ip.Value)
 							}
 						}
 					}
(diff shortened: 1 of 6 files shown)


andrew-sumner:
Heads-up: the failing **Lint & Test** job is unrelated to this change.

The workflow installs `golangci-lint@latest` (now v2.12.2), whose new `inline` govet analyzer reports 20 pre-existing findings across the repo — e.g. `nutanix/common/common.go` (`reflect.Ptr` should be inlined), the v3 SDK `_test.go` files (`ioutil.ReadAll`/`ReadFile`), and existing resource code. It lints the whole repo, not just the diff, so **none of the flagged lines are in this PR** and the same failure occurs on any PR against `master` right now.

GullapalliAkhil:
@andrew-sumner Can you rebase the branch and raise the PR against release/2.5.0? If done on time we are planning to take this in upcoming release.

nutanix/services/vmmv2/resource_nutanix_virtual_machine_v2.go
```diff
@@ -627,6 +628,18 @@ func ResourceNutanixVirtualMachineV2() *schema.Resource {
 				Default:      "ON",
 				ValidateFunc: validation.StringInSlice([]string{"ON", "OFF", "PAUSED", "UNDETERMINED"}, false),
 			},
+			"wait_for_ip_timeout": {
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Default:     5,
```
GullapalliAkhil, line 634: 5 mins is huge?

GullapalliAkhil: Can we have this as constant? instead of hardcode in spec? In future if I want to use the same timeout I can reuse?

nutanix/services/vmmv2/resource_nutanix_virtual_machine_v2_apipa_test.go
```diff
@@ -0,0 +1,60 @@
+package vmmv2
+
+import (
+	"testing"
+
```
GullapalliAkhil, line 5: Can we add this testcase in existing Virtual Machine testcases?

website/docs/r/virtual_machine_v2.html.markdown
```diff
@@ -210,6 +210,8 @@ The following arguments are supported:
 * `gpus`: (Optional) GPUs attached to the VM.
 * `serial_ports`: (Optional) Serial ports configured on the VM.
 * `protection_type`: (Optional) The type of protection applied on a VM. Valid values "PD_PROTECTED", "UNPROTECTED", "RULE_PROTECTED".
+* `wait_for_ip_timeout`: (Optional) Minutes to wait after power-on for the VM to report a usable IPv4 address on its first NIC. A value less than 1 disables the wait. Defaults to `5`. Modeled on the vSphere provider's `wait_for_guest_net_timeout`.
```
GullapalliAkhil, line 213: Can we add the same in the config section?
Can we add the same in Examples as well?

GullapalliAkhil (changes requested):

nutanix/services/vmmv2/resource_nutanix_virtual_machine_v2_apipa_test.go
Haroon-Dweikat-Ntx: this is unit test, can you please add acc test under `nutanix/services/vmmv2/resource_nutanix_virtual_machine_v2_test.go`

Haroon-Dweikat-Ntx (changes requested):

nutanix/services/vmmv2/resource_nutanix_virtual_machine_v2.go
andrew-sumner: Done in d7e6ec34 — both defaults are now package constants next to the existing `timeout`/`delay`:

```go
const (
	timeout = 5 * time.Minute
	delay   = 5 * time.Second

	// defaultWaitForIPTimeoutMinutes is the default for the wait_for_ip_timeout argument,
	// in minutes. It preserves the 5 minute IP wait this resource has always used (see
	// timeout above), and matches the vSphere provider's wait_for_guest_net_timeout.
	defaultWaitForIPTimeoutMinutes = 5

	// defaultWaitForIPRoutable is the default for the wait_for_ip_routable argument:
	// skip APIPA/link-local addresses while waiting for the guest to report an IP.
	defaultWaitForIPRoutable = true
)
```

Also added `TestWaitForIPSchemaDefaults`, which asserts the schema defaults still equal the constants so the two cannot drift.

On "5 mins is huge?" — worth flagging that **5 minutes is not new behaviour introduced by this PR**. The create-time IP wait already used `Timeout: timeout` (the `5 * time.Minute` const above), so the resource has always been willing to wait up to 5 minutes here. This PR only makes that wait *configurable* for the first time; the default was chosen to preserve today's behaviour exactly, and it happens to match vSphere's `wait_for_guest_net_timeout` default.

Also note the wait timing out is non-fatal — it logs `[WARN] could not get the IP for VM(...)` and the resource still completes — so the 5 minutes is a ceiling on a slow/no-DHCP edge case, not a cost paid on a normal create. A VM that gets its lease in 20s stops waiting at 20s. Users who don't want any wait can set `wait_for_ip_timeout = 0`.

Happy to lower the default if you'd prefer, but that would be a behaviour change relative to 2.4.x rather than a preservation of it — let me know which you want.

nutanix/services/vmmv2/resource_nutanix_virtual_machine_v2_unit_test.go
```diff
@@ -0,0 +1,60 @@
+package vmmv2
+
+import (
+	"testing"
+
```
andrew-sumner: Acceptance test added in d7e6ec34 — `TestAccV2NutanixVmsResource_WaitForRoutableIP` in `nutanix/services/vmmv2/resource_nutanix_virtual_machine_v2_test.go`, alongside the other VM acc tests:

- **Step 1** creates a VM with a NIC using the defaults (`wait_for_ip_timeout = 5`, `wait_for_ip_routable = true`) and runs a `testCheckLearnedIPsAreRoutable` state check that walks every `nics.*.…ipv4_info.*.learned_ip_addresses.*.value` in state and fails if any of them is in `169.254.0.0/16` — i.e. it fails on exactly the symptom reported in #871.
- **Step 2** re-applies with `wait_for_ip_timeout = 0` to assert the wait can be disabled without failing the apply.

On the unit tests: I could not fold them into `resource_nutanix_virtual_machine_v2_test.go` because that file (and every other `_test.go` in `nutanix/services/vmmv2/`) is `package vmmv2_test`, and `isAPIPA` / `getFirstIPAddress` are unexported — an external test package can't reach them. Exporting them purely for tests seemed worse than keeping a small internal-package file, so instead I:

- renamed the file `resource_nutanix_virtual_machine_v2_apipa_test.go` → `resource_nutanix_virtual_machine_v2_unit_test.go`, and
- added a header comment stating why it is `package vmmv2` and cross-referencing `TestAccV2NutanixVmsResource_WaitForRoutableIP` as the acceptance coverage for the same feature.

If you'd rather these helpers were exported so everything can live in the one `vmmv2_test` file, say the word and I'll do that instead.

website/docs/r/virtual_machine_v2.html.markdown
andrew-sumner: Done in d7e6ec34 — added in both places:

**`website/docs/r/virtual_machine_v2.html.markdown`** — a new `vm-wait-for-ip` resource in the `## Example` block at the top, showing both arguments with their defaults spelled out and a comment explaining what each one does and how to opt out.

**`examples/virtual_machine_v2/main.tf`** — a new `example-14` resource following the same style as the existing examples (cluster/disk/NIC from the existing data sources), with a comment explaining the APIPA-before-DHCP behaviour this guards against. `terraform fmt` is clean.

The `## Argument Reference` entries for both arguments are unchanged from the original commit.

nutanix/services/vmmv2/resource_nutanix_virtual_machine_v2_unit_test.go
andrew-sumner: Addressed — see the reply on Haroon's follow-up below for the detail. Short version: acceptance coverage (`TestAccV2NutanixVmsResource_WaitForRoutableIP`) now lives in `resource_nutanix_virtual_machine_v2_test.go` with the other VM tests; these unit tests have to stay in a `package vmmv2` file because they call the unexported `isAPIPA`/`getFirstIPAddress` directly, so the file is renamed to `..._unit_test.go` with a comment explaining the split.

andrew-sumner:
@GullapalliAkhil Done — the branch is rebased onto `release/2.5.0` and the PR now targets `release/2.5.0` instead of `master`. GitHub reports it as mergeable, and the diff is confined to the 5 files this change owns (nothing dragged in by the rebase).

The rebase had one real conflict worth calling out: `waitForIPRefreshFunc` gained a `ctx` parameter on `release/2.5.0` (and `GetVmById` now takes `ctx` + a request struct). Resolved by keeping the 2.5.0 signature and adding the new argument alongside it — `waitForIPRefreshFunc(ctx, client, vmUUID, routable)`. `go build ./...`, `go vet ./nutanix/services/vmmv2/`, `gofmt` and the unit tests all pass on the rebased branch.

Review feedback is addressed in d7e6ec34, with replies on each thread:

- **Timeout constant** — `defaultWaitForIPTimeoutMinutes` / `defaultWaitForIPRoutable` consts, pinned by `TestWaitForIPSchemaDefaults`. (Also noted on that thread: the 5 minute wait is pre-existing, not introduced here — this PR only makes it configurable.)
- **Acceptance test** — `TestAccV2NutanixVmsResource_WaitForRoutableIP` in `resource_nutanix_virtual_machine_v2_test.go`, which fails if any learned NIC IP is a `169.254.x` address, plus a step covering `wait_for_ip_timeout = 0`.
- **Docs/examples** — both arguments added to the Example config block in the docs and to `examples/virtual_machine_v2/main.tf`.

Ready for another look whenever suits.

PR #1225, opened 2026-08-18
Fix issue 1218: Add categories to Subnets via Terraform

### LINKED ISSUE #1218: Add categories to Subnets and Nutanix Hosts via Terraform
### Description
Currently, the Terraform provider doesn't support assigning categories to Subnet and Nutanix Host resources. Other resources (e.g., VMs) support categories via the categories block, but this isn't available for nutanix_subnet and nutanix_host.

### Request
Please add a categories attribute/block to the nutanix_subnet and nutanix_host resources so categories can be created, updated, and managed via Terraform, consistent with how they're handled on other resource types.

### Use case
Managing category assignments for subnets and hosts entirely through Terraform, without needing manual steps (or API calls) in Prism Central. Security Rules and Nutanix Host Affinity Policies are based around categories, so it would be nice that this can be created, updated and managed via Terraform.

Haroon-Dweikat-Ntx:
Add metadata section to subnet_v2 documentation

- Introduced a new `metadata` block in the `nutanix_subnet_v2` resource to include optional fields such as `owner_reference_id`, `owner_user_name`, `project_reference_id`, `project_name`, and `category_ids`.
- Updated documentation to clarify the purpose and usage of the new `metadata` fields, enhancing user understanding of subnet management.

### DIFF
--- a/website/docs/r/subnet_v2.html.markdown
+++ b/website/docs/r/subnet_v2.html.markdown
@@ -22,6 +22,13 @@ resource "nutanix_subnet_v2" "vlan-112" {
   subnet_type       = "VLAN"
   network_id        = 122
   is_external       = true
+  metadata {
+    owner_reference_id   = "d0f5c9b0-7b3c-4f5a-9d5b-0a1b2c3d4e5f"
+    owner_user_name      = "admin"
+    project_reference_id = "2defe0f5-6e48-4c9b-b07c-bdd2dc004225"
+    project_name         = "default"
+    category_ids         = ["4c9b0f5d-6e48-2def-b07c-bdd2dc004225"]
+  }
   ip_config {
 
     ipv4 {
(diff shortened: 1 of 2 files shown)


website/docs/r/subnet_v2.html.markdown
```diff
@@ -22,6 +22,13 @@ resource "nutanix_subnet_v2" "vlan-112" {
   subnet_type       = "VLAN"
   network_id        = 122
   is_external       = true
+  metadata {
```
GullapalliAkhil, line 25: Can we update the examples as well? Add a specific section on how users can add the categories?

GullapalliAkhil (changes requested):

website/docs/r/subnet_v2.html.markdown
Haroon-Dweikat-Ntx: example updated.

GullapalliAkhil (approved):

